PR #6, opened 2022-11-08
Device manager

arun3528:


src/device/device-management.ts
```diff
@@ -12,7 +12,7 @@ export enum ErrorTypes {
 /**
  * Represents a WCME error, which contains error type and error message.
```
bbaldino, line 13: Don't forget to update the comment as well.

src/device/device-management.ts
```diff
@@ -33,29 +33,87 @@ export type AudioDeviceConstraints = {
...
 
-export type VideoDeviceConstraints = {
-  deviceId?: ConstrainDOMString;
-  width?: ConstrainULong;
-  height?: ConstrainULong;
-  aspectRatio?: ConstrainDouble;
-  frameRate?: ConstrainDouble;
-  facingMode?: ConstrainDOMString;
+// Predefined encoder config mentioned similar to https://api-ref.agora.io/en/voice-sdk/web/4.x/globals.html#videoencoderconfigurationpreset
+export enum StaticVideoEncoderConfig {
```
bbaldino, line 37: What is this used for? Wondering if it makes sense for this to be defined in core.

src/device/device-management.ts
```diff
@@ -33,29 +33,87 @@ export type AudioDeviceConstraints = {
...
+  exact?: number;
+};
+
+// Predefined encoder config mentioned similar to https://api-ref.agora.io/en/voice-sdk/web/4.x/globals.html#videoencoderconfigurationpreset
+export type videoEncoderConfig = {
+  bitRateMin?: number;
+  bitrateMax?: number;
+  frameRate?: number | ConstrainLong;
+  height?: number | ConstrainLong;
+  width?: number | ConstrainLong;
```
bbaldino, line 54: Did we already have both `ConstrainLong` and `ConstrainULong`? Do we need both?

src/device/device-management.ts
```diff
@@ -33,29 +33,87 @@ export type AudioDeviceConstraints = {
...
+
+export type ConstrainLong = {
+  min?: number;
+  max?: number;
+  ideal?: number;
+  exact?: number;
+};
+
+// Predefined encoder config mentioned similar to https://api-ref.agora.io/en/voice-sdk/web/4.x/globals.html#videoencoderconfigurationpreset
+export type videoEncoderConfig = {
```
bbaldino, line 49: This should start with a capital letter.

src/device/device-management.ts
```diff
@@ -33,29 +33,87 @@ export type AudioDeviceConstraints = {
...
+
+export type VideoConstraints = {
+  cameraDeviceId?: string;
+  faceingMode?: ConstrainULong;
+  optimizationMode?: ConstrainULong;
+  encoderConfig?: StaticVideoEncoderConfig | videoEncoderConfig;
+};
+
+// something like this https://api-ref.agora.io/en/video-sdk/web/4.x/globals.html#audioencoderconfigurationpreset
+export enum StaticAudioEncoderConfig {
```
bbaldino, line 66: Same as above.  Also not sure we want to refer to agora docs in these comments... :)

src/device/device-management.ts
```diff
@@ -33,29 +33,87 @@ export type AudioDeviceConstraints = {
...
+  height?: number | ConstrainLong;
+  width?: number | ConstrainLong;
+  aspectRatio?: number;
+};
+
+export type VideoConstraints = {
+  cameraDeviceId?: string;
+  faceingMode?: ConstrainULong;
+  optimizationMode?: ConstrainULong;
+  encoderConfig?: StaticVideoEncoderConfig | videoEncoderConfig;
```
bbaldino, line 62: So fields were removed from this?  I think this was designed to mimic the underlying API.

src/device/device-management.ts
bbaldino: Even if we do have this, instead of being it's own type, we should just define named, constant instances of `VideoEncoderConfig` that have the fields pre-set.

src/device/device-management.ts
```diff
@@ -33,29 +33,87 @@ export type AudioDeviceConstraints = {
...
+};
+
+// /**
+//  * @param audioConfig
+//  * @param videoConfig
+//  */
+// export async function createMicrophoneAndCameraTracks(
+//   audioConfig?: MicrophoneConstraints,
+//   videoConfig?: VideoConstraints
+// ): Promise<[IMicrophoneAudioTrack, ICameraVideoTrack]> {}
```
bbaldino, line 102: Should either uncomment or remove this.

src/device/device-management.ts
```diff
@@ -33,29 +33,87 @@ export type AudioDeviceConstraints = {
...
  */
-export async function createCameraTrack(
-  constraints?: VideoDeviceConstraints
-): Promise<LocalCameraTrack> {
+export async function createCameraTrack(constraints?: VideoConstraints): Promise<LocalCameraTrack> {
   let stream: MediaStream;
   try {
-    stream = await media.getUserMedia({ video: { ...constraints } });
+    // TODO: change it
+    stream = await media.getUserMedia({ video: {} });
```
bbaldino, line 114: This should be using `constraints`?

src/device/device-management.ts
```diff
@@ -70,30 +128,60 @@ export async function createCameraTrack(
...
  * Creates a display video track.
  *
+ * @param constraints
  * @returns A Promise that resolves to a LocalDisplayTrack.
  */
-export async function createDisplayTrack(): Promise<LocalDisplayTrack> {
+export async function createDisplayTrack(
+  constraints?: VideoConstraints
+): Promise<LocalDisplayTrack> {
   const stream = await media.getDisplayMedia({ video: true });
```
bbaldino, line 155: Should be using `constraints`?

src/device/device-management.ts
```diff
@@ -70,30 +128,60 @@ export async function createCameraTrack(
...
   return new LocalDisplayTrack(stream);
 }
 
+// /**
+//  * Creates custom video track .
+//  *
+//  * @param constraints
+//  * @returns A Promise that resolves to a LocalVideoTrack.
+//  */
```
bbaldino, line 164: Remove or uncomment all this?

src/device/device-management.ts
```diff
@@ -119,7 +210,7 @@ export async function getDevices(deviceKind?: media.DeviceKind): Promise<MediaDe
  *
  * @returns List of microphone devices in an array of MediaDeviceInfo objects.
  */
-export async function getAudioInputDevices(): Promise<MediaDeviceInfo[]> {
+export async function getMicrophones(): Promise<MediaDeviceInfo[]> {
```
bbaldino, line 213: I kinda feel like webrtc-core should use the vocabulary of the underlying API (devices).  (Here and other naming above)

bbaldino (commented):
there's a lot of semantically different changes in here that I think would be nice to split into separate PRs.

arun3528:
Thanks @bbaldino more changes to come, would not recommend to review the PR for now , Will update you once ready

src/device/device-management.ts
arun3528: Will clean up the comments but the staticVideoEncoderConfig is a enum the developers would have to do 

constrains.encoderConfig.preDefinedEncoderConfig compared to constraints.encoderConfig 

is that the ask !?

src/device/device-management.ts
arun3528: No we dont removing the constainULong

src/device/device-management.ts
arun3528: done

src/device/device-management.ts
arun3528: yes we need to generate the constraints , @haohancom  is going to add it

src/device/device-management.ts
arun3528: we could but again the SDK have to add a wrapper to simply map the function name , simply more work , 

as the name is more internal can we re consider it

src/device/device-management.ts
bbaldino: No, what I'm saying is this:

rather than having two types, e.g.:
```
export type VideoEncoderConfig {
   maxHeight: number,
   maxWidth: number,
   maxFps: number,
   // etc (just an example)
}
export enum StaticVideoEncoderConfig {
    1080p,
    720p,
    360p
}
```
do:
```
export type VideoEncoderConfig {
   maxHeight: number,
   maxWidth: number,
   maxFps: number,
   // etc (just an example)
}

export const highQualityVideoEncoiderConfig: VideoEncoderConfig = {
   maxHeight: 1080,
   maxWidth: 1920,
   maxFps: 30,
}
```

So: a single type that is always used, and pre-defined settings are just defined as constant values of that type, so that way the user can have convenience if it wants, and the code only has to deal with a single type.

PR #7, opened 2022-11-10
chore: make custom config work for audio and video 

haohancom:


src/device/device-management.ts
```diff
@@ -86,7 +87,16 @@ export async function createMicrophoneAndCameraTracks(
...
+    if (
+      constraints?.encoderConfig &&
+      typeof constraints.encoderConfig === 'string' &&
+      encoderConfigMap.has(constraints.encoderConfig)
+    ) {
+      // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+      // @ts-ignore
+      // eslint-disable-next-line no-param-reassign
+      constraints.encoderConfig = encoderConfigMap.get(constraints.encoderConfig);
+    }
```
bbaldino, line 99: @arun3528 this is the kind of logic that would be cleaner if we use a constant value for the static configs instead of a separate type.

src/device/device-management.ts
```diff
@@ -86,7 +87,16 @@ export async function createMicrophoneAndCameraTracks(
...
     stream = await media.getUserMedia({ video: {} });
```
haohancom, line 100: @arun3528 , do we need to convert constraints into video object here?

src/device/device-management.ts
arun3528: yep going to change it

src/device/device-management.ts
arun3528: yes we will have to

src/device/device-management.ts
arun3528: @haohancom  you dont have to do any checks on the code here the developers will directly pass the encoder config 

We will just export a constant which developer can choose one and pass it in 

```
export const staticVideoEncoderConfig = [
  '1080P',
  '720P',
  '480P',
  '360P',
  '240P',
  '180P',
  '120P',
] as const;

type StaticVideoEncoderConfig = typeof staticVideoEncoderConfig[number];

export const predefinedVideoEncoderConfig: Record<StaticVideoEncoderConfig, VideoEncoderConfig> = {
  '1080P': { frameRate: 15, bitrateMax: 2080 },
  '720P': { frameRate: 15, bitrateMax: 1130 },
  '480P': { frameRate: 15, bitrateMax: 500 },
  '360P': { frameRate: 15, bitrateMax: 400 },
  '240P': { frameRate: 15, bitrateMax: 200 },
  '180P': { frameRate: 15, bitrateMax: 140 },
  '120P': { frameRate: 15, bitrateMax: 65 },
} as const;
```
Developers will just do

createCameraTrack({
  cameraDeviceId: '';
  encoderConfig: predefinedVideoEncoderConfig.1080P;
})

@bbaldino  this is what you want correct. ?

bbaldino: I guess this kinda does what I had in mind, but I still think it kinda over complicates things.  I was thinking of just something like:
```
export const PredefinedVideoEncoderConfig1080p = { frameRate: 15, bitrateMax: 2080 }
export const PredefinedVideoEncoderConfig720p = { frameRate: 15, bitrateMax: 1130 }
etc.
```
So we don't need the separate `staticVideoEncoderConfig` enum, and we don't need `Record<StaticVideoEncoderConfig, VideoEncoderConfig>`, you just have one type: `VideoEncoderConfig` and users can either fill that out manually, or use a pre-defined constant value of that type that corresponds to a pre-defined config (`PredefinedVideoEncoderConfig1080p`, `PredefinedVideoEncoderConfig720p`, etc.)

arun3528: you are correct that the encoderconfig should have one type 

but the pre defined config , if its 2 or 3 we can export them individually but if its more its better to export them as a constant so that developers get to know all the config using the typescript  

i am in for both the options we can change if we need to we dont need record we could further simplify it

bbaldino: If I'm understanding you correctly, we could do something like:
```
export type StaticVideoConfig = {
    1080p: { frameRate: 15, bitrateMax: 2080 },
    720p: { frameRate: 15, bitrateMax: 1130 },
}
```
or with static members of a class maybe?
```
class StaticVideoConfig {
    static 1080p: VideoEncoderConfig = { frameRate: 15, bitrateMax: 2080 };
    ...
}
```

something like that if you want to group them.

arun3528: like the first one if we need to change in the future we can

samples/app.js
```diff
@@ -35,7 +36,7 @@ function handleError(error) {
 function buildSelectOption(device) {
   const childOption = document.createElement('option');
 
-  childOption.value = device.ID;
+  childOption.value = device.deviceId;
```
haohancom, line 39: Fix `gather devices` button can not get correct camera devices issue.

PR #8, opened 2022-11-11
feat: added connection state handling

marcin-bazyl:
Added connection state handler that takes ice connection state and dtls connection state and calculates the final overall state from these two.

Also fixed various warnings that I've found are currently on master.

### DIFF
--- a/src/connection-state-handler.ts
+++ b/src/connection-state-handler.ts
@@ -1,0 +1,118 @@
+import { EventEmitter, EventMap } from './event-emitter';
+import { logger } from './util/logger';
+
+// Overall connection state (based on the ICE and DTLS connection states)
+export enum ConnectionState {
+  New = 'New', // connection attempt has not been started
+  Closed = 'Closed', // connection closed, there is no way to move out of this state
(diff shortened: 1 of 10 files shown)


src/connection-state-handler.ts
```diff
@@ -0,0 +1,116 @@
...
+   * Handler for ice connection state change.
+   */
+  public onIceConnectionStateChange(): void {
+    const newConnectionState = this.evaluateMediaConnectionState();
+
+    if (newConnectionState !== this.mediaConnectionState) {
+      this.mediaConnectionState = newConnectionState;
+      this.emit(ConnectionStateEvents.ConnectionStateChanged, this.mediaConnectionState);
+    }
+  }
```
bbaldino, line 72: It looks like this and `onConnectionStateChange` could both leverage some private helper method?

bbaldino (approved):
one small nit, but not a huge deal.

src/connection-state-handler.ts
marcin-bazyl: done

bbaldino (approved):

PR #13, opened 2022-11-30
chore:  changes device manager functionality

arun3528:
1) changes some name for the api 
2) adds creation of custom tracks if someone wants to add it 
3) adds static config for track creation


src/device/device-management.ts
```diff
@@ -70,28 +103,67 @@ export async function createCameraTrack(
...
+ * @param constraints - Display constraints for screen sharing.
  * @returns A Promise that resolves to a LocalDisplayTrack.
  */
-export async function createDisplayTrack(): Promise<LocalDisplayTrack> {
-  const stream = await media.getDisplayMedia({ video: true });
-  return new LocalDisplayTrack(stream);
+export async function createDisplayTrack(
+  constraints?: VideoConstraints
+): Promise<LocalDisplayTrack> {
+  const stream = await media.getDisplayMedia({ video: generateVideoConstraints(constraints) });
```
marcin-bazyl, line 130: we probably need a different type for display track constraints argument (and different method for generating them from the supplied arg), because browsers have some specific things for screen sharing, for example see here: https://developer.chrome.com/docs/web-platform/screen-sharing-controls/ for recent additions in Chrome to getDisplayMedia constraints - we should expose them to the clients

src/device/device-management.ts
```diff
@@ -70,28 +103,67 @@ export async function createCameraTrack(
...
+  return new LocalDisplayTrack(stream.getVideoTracks()[0]);
+}
+
+/**
+ * Creates custom video track .
+ *
+ * @param config - Custom video track creation config.
+ * @returns A Promise that resolves to a LocalVideoTrack.
+ */
+export async function createCustomVideoTrack(
```
marcin-bazyl, line 140: not sure if we actually need these "createCustom*Track" functions... why not just add encoder config as a parameter to the LocalVideoTrack/LocalAudioTrack constructor?

src/device/device-management.ts
```diff
@@ -29,38 +36,64 @@ export class WcmeError {
...
+ * @param videoConstraints - Video constraints to create camera track.
+ * @returns MicrophoneTrack and cameraTrack at same time.
+ */
+export async function createMicrophoneAndCameraTracks(
+  audioConstraints?: MicrophoneConstraints,
+  videoConstraints?: VideoConstraints
+): Promise<[LocalMicrophoneTrack, LocalCameraTrack]> {
+  let stream: MediaStream;
+  try {
+    // TODO: change it
```
bbaldino, line 62: change what? i'd either get rid of the TODO or add more detail

src/device/device-management.ts
```diff
@@ -29,38 +36,64 @@ export class WcmeError {
...
  * @returns A LocalTrack object or an error.
  */
-export async function createCameraTrack(
-  constraints?: VideoDeviceConstraints
-): Promise<LocalCameraTrack> {
+export async function createCameraTrack(constraints?: VideoConstraints): Promise<LocalCameraTrack> {
   let stream: MediaStream;
   try {
-    stream = await media.getUserMedia({ video: { ...constraints } });
+    // TODO: change it
```
bbaldino, line 88: Same as above

src/media/utils.ts
```diff
@@ -0,0 +1,56 @@
...
+};
+
+/**
+ * Generates the audio constraints pass to getUserMedia.
+ *
+ * @param videoConstraints - Video constraints.
+ * @returns Native video constraints.
+ */
+export const generateVideoConstraints = (
+  videoConstraints?: VideoConstraints
```
bbaldino, line 43: the extra `VideoConstraints` type doesn't really seem to add much value, if we really want a separate type i'd use a `Partial` of the `MediaTrackConstraints` for video and audio

src/device/device-management.ts
```diff
@@ -29,38 +36,64 @@ export class WcmeError {
...
+export async function createMicrophoneAndCameraTracks(
+  audioConstraints?: MicrophoneConstraints,
+  videoConstraints?: VideoConstraints
+): Promise<[LocalMicrophoneTrack, LocalCameraTrack]> {
+  let stream: MediaStream;
+  try {
+    // TODO: change it
+    stream = await media.getUserMedia({
+      audio: generateAudioConstraints(audioConstraints),
+      video: generateVideoConstraints(videoConstraints),
```
bbaldino, line 65: See comment elsewhere on the need for the `VideoConstraints` and `MicrophoneConstraints` types and the need for `generateVideoConstraints` and `generateAudioConstraints`

src/device/device-management.ts
```diff
@@ -70,28 +103,67 @@ export async function createCameraTrack(
...
+ * @param config - Custom video track creation config.
+ * @returns A Promise that resolves to a LocalVideoTrack.
+ */
+export async function createCustomVideoTrack(
+  config: CustomVideoTrackInitConfig
+): Promise<LocalVideoTrack> {
+  const localVideoTrack = new LocalVideoTrack(config.mediaStreamTrack);
+
+  if (config.encoderConfig) {
+    localVideoTrack.setEncoderConfig(config.encoderConfig);
```
bbaldino, line 146: where is `setEncoderConfig` defined? i don't see it in this PR and i don't see it in the `new-tracks` branch.

src/device/device-management.ts
```diff
@@ -29,38 +36,64 @@ export class WcmeError {
...
+  let stream: MediaStream;
+  try {
+    // TODO: change it
+    stream = await media.getUserMedia({
+      audio: generateAudioConstraints(audioConstraints),
+      video: generateVideoConstraints(videoConstraints),
+    });
+  } catch (error) {
+    throw new WebrtcError(
+      ErrorTypes.CREATE_CAMERA_TRACK_FAILED,
```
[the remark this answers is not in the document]
arun3528: But the errors on thrown on the createMicrophoneTrack and createVideoTrack

src/device/device-management.ts
arun3528: i agree can do that in a separate PR

src/device/device-management.ts
arun3528: This will be more of a developer facing function , good to be more specific, If it was a internal class would have gone with the constructor way

src/device/device-management.ts
arun3528: its in a different PR let me consolidate them

src/media/utils.ts
arun3528: it wont work because we are giving custom names to them microphoneDeviceId, cameraDeviceId, encoderConfig

src/device/device-management.ts
arun3528: When someone passes a values they should know if the constraints are for audio or video having specific keys to the types makes it easy for developers to understand 

just passing in MediaTrackConstraints is easy internally but when it comes to types better to be more specific 
there is also new constaints coming in for displayMedia which is also the same constaints

src/device/device-management.ts
bbaldino: I agree with @marcin-bazyl , I don't actually think this is friendlier: it ends up being more confusing in my opinion because they aren't really that different, and the alternate API makes it more confusing as to which should be used.

src/media/utils.ts
bbaldino: In my opinion those names don't add value that make it worthwhile to introduce the burden of maintaining the extra types.  I don't think `webrtc-core` needs to strive to be friendly to people who aren't familiar with webrtc details, instead it should just be convenient for those who are familiar to accomplish common tasks.

src/device/device-management.ts
bbaldino: we can still have aliases, but should be based on `Partial`s of the existing types.

bbaldino: we can still have aliases, but should be based on `Partial`s of the existing types.

src/media/utils.ts
arun3528: In that case the meetings SDK will need to add these type in there code for the developers 

just wanted to avoid adding those types in calling and meetings SDK rather do it here we can change it if you are ok with it

arun3528: let me check with marcin on this and get back

src/device/device-management.ts
arun3528: Was looking into this @marcin-bazyl  , When someone passes the track we can ask them to do 

let customTrack = new LocalVideoTrack(track)
customTrack.setEncoderConfig(config)

or 

let customTrack = new LocalVideoTrack(track, encoderConfig)

which will indirectly save and apply the encoderConfig

arun3528: I removed those function now @bbaldino  @marcin-bazyl

src/device/device-management.ts
arun3528: done

arun3528:
@bbaldino @marcin-bazyl have remove the custom track and used pick to pick types from mediaStreamConstraints

src/device/device-management.ts
arun3528: to confirm will add to a separate PR

src/device/device-management.ts
arun3528: done

src/device/device-management.ts
arun3528: removed

src/media/utils.ts
arun3528: moved to start using the pick type

src/device/device-management.ts
arun3528: removed this

src/media/local-video-track.ts
```diff
@@ -10,46 +10,33 @@ export enum OptimizationMode {
...
 
-export type VideoEncoderConfig = {
-  bitRateMin?: number;
-  bitrateMax?: number;
-  frameRate?: number | ConstrainLong;
-  height?: number | ConstrainLong;
-  width?: number | ConstrainLong;
-  aspectRatio?: number;
-};
+export interface VideoConstraints extends BaseVideoConstraints {
```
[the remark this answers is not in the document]
arun3528: makes sense we are anyways going to add one more for share so we can rename it

src/media/track.ts
```diff
@@ -0,0 +1,223 @@
...
+  [Events.UnderlyingTrackChange]: () => void;
+}
+
+export type TrackEffect = any;
+// TBD: Fix this once types are published seperatly
+// export type TrackEffect = BaseMicrophoneEffect | BaseCameraEffect;
+
+/**
+ * Basic Track class. Wrapper for LocalTrack from 'webrtc-core'.
+ */
```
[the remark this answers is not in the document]
arun3528: fixed

src/media/track.ts
```diff
@@ -0,0 +1,223 @@
...
+   * Constructor for the Track class. Creates an empty CoreLocalTrack or uses an existing one.
+   *
+   * @param stream - The MediaStream for this LocalTrack.
+   * @param track
+   */
+  constructor(track: MediaStreamTrack) {
+    super();
+    /**
+     * Emit ended event when the underlying track ends.
+     */
```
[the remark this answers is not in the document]
arun3528: done

src/device/device-management.ts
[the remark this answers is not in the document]
bbaldino: is this resolved?

src/media/track.ts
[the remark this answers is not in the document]
bbaldino: This comment still appears to be wrong? It says this is a "wrapper for `LocalTrack` in 'webrtc-core', but that isn't the case.

src/device/device-management.ts
```diff
@@ -109,7 +109,7 @@ export async function createMicrophoneTrack(
  * @returns A Promise that resolves to a LocalDisplayTrack.
  */
 export async function createDisplayTrack(
-  constraints?: VideoConstraints,
+  constraints?: CameraConstraints,
```
bbaldino, line 112: Sorry, maybe I missed this when I made my earlier comment, but if the constraints here are generic for all video (camera and display) we should probably stick with `VideoConstraints`.  Same for audio: if they apply to both microphone and 'computer', then it should just be `AudioConstraints`.

If there _are_ different constraints for camera/display and microphone/computer, then we should have 4 different constraint types.

src/media/local-track.ts
```diff
@@ -34,7 +34,7 @@ export type TrackEffect = any;
 // export type TrackEffect = BaseMicrophoneEffect | BaseCameraEffect;
 
 /**
- * Basic Track class. Wrapper for LocalTrack from 'webrtc-core'.
+ * Local track manages effects which gets extended by video and audio.
```
bbaldino, line 37: Oops, I must have had a filter on when I saw this line before (I didn't see the updated comment).

src/device/device-management.ts
arun3528: we are going to add a new constrains for display later so it is better to have microphone and camera previoly we were using one which does not make sense now as share has other parameters compared to camera

src/device/device-management.ts
marcin-bazyl: we should have a jira for this under your [epic](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-375070)

src/device/device-management.ts
```diff
@@ -108,7 +139,10 @@ export async function getDevices(deviceKind?: media.DeviceKind): Promise<MediaDe
       media.enumerateDevices
     );
   } catch (error) {
-    throw new WcmeError(ErrorTypes.DEVICE_PERMISSION_DENIED, 'Failed to ensure device permissions');
+    throw new WebrtcError(
```
marcin-bazyl, line 142: shouldn't we be throwing this error directly from `media.ensureDevicePermissions`?

src/media/local-display-track.ts
```diff
@@ -1,6 +1,6 @@
-import { LocalTrack } from './local-track';
+import { LocalVideoTrack } from './local-video-track';
 
 /**
  * Represents a local track for a display source.
  */
-export class LocalDisplayTrack extends LocalTrack {}
+export class LocalDisplayTrack extends LocalVideoTrack {}
```
marcin-bazyl, line 6: as you're adding LocalDisplayTrack in this PR, can you also add the LocalComputerAudioTrack too?

src/media/local-audio-track.ts
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Represents a audio track for a audio source.
+ */
+export class LocalAudioTrack extends LocalTrack {
+  /**
+   * Makes sure to apply the encoderConfig for the audio.
+   *
+   * @param encoderConfig - Encoder config for audio.
+   */
+  setEncoderConfig(encoderConfig: MicrophoneConstraints): void {
```
marcin-bazyl, line 25: not sure about the name - the function name and argument use "EncoderConfig" name, but the type is "Constraints", we should be consistent.

src/media/local-video-track.ts
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Represents a local video track.
+ */
+export class LocalVideoTrack extends LocalTrack {
+  /**
+   * Applies the passed video constraints.
+   *
+   * @param constraints - Custom encoder config for video.
+   */
+  setEncoderConfig(constraints: CameraConstraints): void {
```
marcin-bazyl, line 56: same thing about the names as in my previous comment "econder config" vs "constraints"

src/media/track.ts
```diff
@@ -0,0 +1,220 @@
...
+   * pass the video element dom element where we need to attach the stream.
+   */
+
+  /**
+   * @param element
+   */
+  play(element?: string | HTMLElement): void {
+    if (element instanceof HTMLElement) {
+      // eslint-disable-next-line no-param-reassign
+      element.srcObject = new MediaStream([this.#mediaStreamTrack]);
```
marcin-bazyl, line 212: if we have effects applied, the srcObject should probably use a stream constructed from `this.#mediaStreamTrackWithEffect` track, right? (same in line 216)

src/device/device-management.ts
```diff
@@ -10,9 +12,9 @@ export enum ErrorTypes {
 }
 
 /**
- * Represents a WCME error, which contains error type and error message.
+ * Represents a webrtc core error, which contains error type and error message.
  */
-export class WcmeError {
+export class WebrtcError {
```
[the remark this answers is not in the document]
arun3528: raised a ticket here https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-375070

src/device/device-management.ts
```diff
@@ -70,28 +86,43 @@ export async function createCameraTrack(
...
+    stream = await media.getUserMedia({ audio: constraints });
   } catch (error) {
-    throw new WcmeError(
+    throw new WebrtcError(
       ErrorTypes.CREATE_MICROPHONE_TRACK_FAILED,
       `Failed to create microphone track ${error}`
     );
   }
-  return new LocalMicrophoneTrack(stream);
+  // See if we can just pass the track and not streams
```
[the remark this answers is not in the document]
arun3528: removed , we have already migrated to pass track

src/device/device-management.ts
```diff
@@ -70,28 +86,43 @@ export async function createCameraTrack(
...
  */
-export async function createDisplayTrack(): Promise<LocalDisplayTrack> {
-  const stream = await media.getDisplayMedia({ video: true });
-  return new LocalDisplayTrack(stream);
+export async function createDisplayTrack(
+  constraints?: CameraConstraints,
+  withAudio?: boolean
+): Promise<{
+  localDisplayTrack: LocalDisplayTrack;
+  localAudioTrack?: LocalAudioTrack | undefined;
```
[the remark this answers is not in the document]
arun3528: done adding the constraints for displayTrack

src/media/local-display-track.ts
arun3528: i thought we agreed on using the LocalAudio Track as it is the same there are no extra property which are different from localAudioTrack

src/device/device-management.ts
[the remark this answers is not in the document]
arun3528: fixed it

src/media/local-audio-track.ts
arun3528: ya previouly we had a different type now its different

src/media/local-track.ts
```diff
@@ -52,77 +34,46 @@ export type TrackEffect = any;
...
   /**
-   * Get muted state of this track.
-   *
-   * @returns The muted state of this track.
+   * Disposes the effect already applied .
    */
-  get muted(): boolean {
-    return !this.underlyingTrack.enabled;
+  disposeEffects(): void {
+    this.effects.forEach((effect: TrackEffect) => effect.dispose());
```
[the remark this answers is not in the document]
arun3528: still have to flushout the events for effects , will do it after i test it out i have to  clear the api first before going into effect

src/media/local-video-track.ts
```diff
@@ -0,0 +1,59 @@
...
+export interface CameraConstraints extends BaseVideoConstraints {
+  facingMode?: FacingMode;
+  optimizationMode?: OptimizationMode;
+}
+
+/**
+ * A class to map resolution with video constraints.
+ */
+export const StaticVideoEncoderConfig: { [key: string]: CameraConstraints } = {
+  '1080p': { frameRate: 15, width: 1920, height: 1080 },
```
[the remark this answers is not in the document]
arun3528: ya thats correct for share its less , done changing

src/media/local-video-track.ts
arun3528: done

src/media/track.ts
```diff
@@ -0,0 +1,220 @@
...
+    super();
+    this.ID = track.id;
+    this.status = track.readyState as TrackStatus;
+    this.label = track.label;
+    this.#mediaStreamTrack = track;
+    this.isPlaying = false;
+    /**
+     * Emit ended event when the underlying track ends.
+     */
+    this.#mediaStreamTrack.onended = () => {
```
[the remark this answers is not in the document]
arun3528: created a ticket for this https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-398390

src/media/track.ts
```diff
@@ -0,0 +1,220 @@
...
+
+    /**
+     * Emit mute event when the underlying track gets muted.
+     */
+    this.#mediaStreamTrack.onmute = () => {
+      // using arrow function which should bind to this from outer scope track
+      const action = this.#mediaStreamTrack.enabled ? 'muted' : 'unmuted';
+      // TODO:  Move this logic else where
+
+      this.emit('track:mute', {
```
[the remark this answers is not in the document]
arun3528: done

src/media/track.ts
```diff
@@ -0,0 +1,220 @@
...
+    this.#mediaStreamTrack.onended = () => {
+      this.emit(Events.Ended, { trackState: this.trackState });
+    };
+
+    /**
+     * Emit mute event when the underlying track gets muted.
+     */
+    this.#mediaStreamTrack.onmute = () => {
+      // using arrow function which should bind to this from outer scope track
+      const action = this.#mediaStreamTrack.enabled ? 'muted' : 'unmuted';
```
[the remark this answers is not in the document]
arun3528: i had always this confusion but on local muted enabled is the values we need to check 

but the recommendation is to only check it onmuted and onUnMuted

src/media/track.ts
```diff
@@ -0,0 +1,220 @@
...
+     * Emit ended event when the underlying track ends.
+     */
+    this.#mediaStreamTrack.onended = () => {
+      this.emit(Events.Ended, { trackState: this.trackState });
+    };
+
+    /**
+     * Emit mute event when the underlying track gets muted.
+     */
+    this.#mediaStreamTrack.onmute = () => {
```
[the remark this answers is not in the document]
arun3528: will check it with other events as part of the effects changes

src/media/track.ts
```diff
@@ -0,0 +1,220 @@
...
+    this.emit(Events.UnderlyingTrackChange);
+  }
+
+  /**
+   * Stop this track.
+   *
+   * @fires Events.Ended
+   */
+  stop(): void {
+    this.#mediaStreamTrack.stop();
```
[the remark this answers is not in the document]
arun3528: we will have to do more if someone does stop , dispose close clear everything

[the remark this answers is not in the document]
arun3528: noted it here https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-398390

src/media/track.ts
arun3528: added to https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-398390

src/media/local-track.ts
```diff
@@ -52,77 +34,46 @@ export type TrackEffect = any;
...
+   * Adds an effect to a local track.
    *
-   * @returns The id of this track.
+   * @param name - The name of the effect.
+   * @param effect - The effect to add.
    */
-  get id(): string {
-    return this.underlyingTrack.id;
+  async addEffect(name: string, effect: TrackEffect): Promise<void> {
+    await effect.load(new MediaStream([this.getMediaStreamTrack()]));
```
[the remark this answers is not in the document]
arun3528: added this too https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-398390

src/media/local-track.ts
```diff
@@ -39,11 +24,8 @@ export type TrackPublishEvent = {
   trackState: TrackState;
 };
 
-export interface TrackEvents extends EventMap {
-  [LocalTrackEvents.Ended]: (event: TrackEndEvent) => void;
-  [LocalTrackEvents.Muted]: (event: TrackMuteEvent) => void;
-  [LocalTrackEvents.PublishedStateUpdate]: (event: TrackPublishEvent) => void;
-  [LocalTrackEvents.UnderlyingTrackChange]: () => void;
+export interface LocalTrackEvents extends EventMap {
+  [Events.PublishedStateUpdate]: (event: TrackPublishEvent) => void;
```
bbaldino, line 28: Don't we still want the other events?

src/media/local-video-track.ts
```diff
@@ -0,0 +1,75 @@
...
+  MediaTrackConstraints,
+  'aspectRatio' | 'height' | 'width' | 'frameRate' | 'deviceId'
+>;
+
+type BaseDisplayConstraints = Pick<
+  MediaTrackConstraints,
+  'aspectRatio' | 'height' | 'width' | 'frameRate' | 'suppressLocalAudioPlayback'
+>;
+
+export interface CameraConstraints extends BaseVideoConstraints {
```
bbaldino, line 29: Is `BaseVideoConstraints` used anywhere else? Looks like it's not needed (same for `BaseDisplayConstraints`) and `CameraConstraints` and `DisplayConstraints` can just be defined from the constraints directly?

src/media/local-track.ts
bbaldino: Ah, I see they moved up to `Track`

bbaldino: I think `UnderlyingTrackChange` will only apply for local tracks?

src/media/track.ts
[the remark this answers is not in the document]
bbaldino: ping ^

src/media/track.ts
```diff
@@ -0,0 +1,220 @@
...
+// TBD: Fix this once types are published seperatly
+// export type TrackEffect = BaseMicrophoneEffect | BaseCameraEffect;
+
+/**
+ * Basic Track class. Wrapper for LocalTrack from 'webrtc-core'.
+ */
+export abstract class Track extends EventEmitter<TrackEvents> {
+  static Events = Events;
+
+  isPlaying: boolean;
```
bbaldino, line 69: Is this needed? The track state is already modeled by the underlying track, no?

src/media/track.ts
```diff
@@ -0,0 +1,220 @@
...
+  /**
+   *
+   * @param element - Either pass the ID for the dom element where we need to attach the video tag
+   * pass the video element dom element where we need to attach the stream.
+   */
+
+  /**
+   * @param element
+   */
+  play(element?: string | HTMLElement): void {
```
bbaldino, line 209: Does this need to be part of the track API?  I thought we were leaving this to the user.

src/media/track.ts
bbaldino: I see now this refers to the element, see my comment on that below.

src/media/local-track.ts
arun3528: the underlying track update will be on the track class , which is the one saving the local and track with effect

src/media/local-video-track.ts
arun3528: pick does not work with extends , so i have to define it before adding other values to the interface

@marcin-bazyl  had the same question the only way is to do 

`export interface CameraConstraints extends xxx` now

src/media/track.ts
arun3528: This to if someone does a play and pass src element we keep the status if its playing actively

src/media/local-track.ts
bbaldino: > the underlying track update will be on the track class , which is the one saving the local and track with effect

sorry, don't think I understand.  `UnderlyingTrackChange` only applies to local tracks, and therefore should be defined at the `LocalTrack` level, not the `Track` level, right?

src/media/track.ts
arun3528: This would be really nice to have so that we can keep track on the dom element where the audio or video is playing 

mainly in the mobile device we can monitor the status of audio and video getting played for metrics and also see if its working properly , we could also play if underlying track changes 

its is more of a helper else developer have to write there own custom logic

bbaldino: Could it be done at a higher level then? It's not something that feels like belongs in webrtc-core.

arun3528: i can add them on just remote track later

bbaldino: they should definitely be on remote track if we're going to keep this api, but i'm talking about moving this functionality out of webrtc-core altogether

arun3528: When thinking about it , i might have to extend all the classes like camera , microphone , display methods

let me see if there is any other way i can introduce this function

arun3528: removed it

src/media/local-track.ts
arun3528: thats true, The main problem is the track change action can happen from localCameraTrack and local-track for effect

which indirectly calls the track to change the underlying track and switch it , moving it to local-track, i might have to move all the effects back to local-track which will cause remoteTrack to not have effect 

i dont think its going to make much different if are making sure the change event is getting triggered from one place

src/device/device-management.ts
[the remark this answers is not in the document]
marcin-bazyl: it's not entirely fixed as the class doens't inherit from Error and we still have the enum, but I'm assuming these things will be taken care of in [SPARK-375070](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-375070)

src/device/device-management.ts
```diff
@@ -101,15 +109,14 @@ export async function createDisplayTrack(): Promise<LocalDisplayTrack> {
...
+    .ensureDevicePermissions(
       [media.DeviceKind.AudioInput, media.DeviceKind.VideoInput],
       media.enumerateDevices
-    );
-  } catch (error) {
-    throw new WcmeError(ErrorTypes.DEVICE_PERMISSION_DENIED, 'Failed to ensure device permissions');
-  }
+    )
+    .catch((error) => {
+      throw error;
```
marcin-bazyl, line 118: what's the point of this catch here now? I think it can be removed

src/media/local-display-track.ts
marcin-bazyl: I think the issue is in WCME that it needs to detect if a published track is a microphone or screen share audio one - it's doing it using `instanceof` and so we need a separate class, because both mic and screen share audio would return true for `instanceof LocalAudioTrack`

src/media/track.ts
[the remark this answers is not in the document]
marcin-bazyl: but we're inside onmute here, so why are we checking `enabled`?

src/media/track.ts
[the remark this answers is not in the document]
marcin-bazyl: ok, so all the effect related issues are in that jira and I've just noticed that this PR is not for master but "new-tracks" branch, so the assumption is that all effects related stuff will be done in another PR for "new-tracks" branch and then altogether will be merged to master, correct?

src/device/device-management.ts
```diff
@@ -1,66 +1,60 @@
...
-}
+import { ErrorTypes, WebrtcError } from '../error/WebrtcError';
 
 /**
- * Represents a WCME error, which contains error type and error message.
+ * Creates MicrophoneTrack and CameraTrack at the same time.
+ *
+ * @param audioConstraints - Audio constraints to create microphone track.
+ * @param videoConstraints - Video constraints to create camera track.
+ * @returns MicrophoneTrack and cameraTrack at same time.
```
marcin-bazyl, line 15: this should say LocalMicrophoneTrack and LocalCameraTrack (same for comments in line 11)

src/device/device-management.ts
arun3528: done

src/device/device-management.ts
arun3528: done

src/device/device-management.ts
```diff
@@ -1,66 +1,65 @@
...
+import { ErrorTypes, WebrtcError } from '../error/WebrtcError';
 
 /**
- * Represents a WCME error, which contains error type and error message.
+ * Creates a camera video track. Please note that the constraint params in second getUserMedia call would NOT take effect when:
+ *
+ * 1. Previous captured video track from the same device is not stopped .
+ * 2. Previous createCameraTrack() call for the same device is in progress.
+ * Creates MicrophoneTrack and CameraTrack at the same time.
+ * Creates LocalMicrophoneTrack and LocalCameraTrack at the same time.
```
marcin-bazyl, line 16: this line is duplicate with line 15, line 15 should be removed

arun3528: removed

src/device/device-management.ts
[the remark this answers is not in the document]
arun3528: ya moved it

src/device/device-management.ts
arun3528: created display track here

src/media/track.ts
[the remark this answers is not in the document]
arun3528: removed

src/device/device-management.ts
arun3528: done

src/device/device-management.ts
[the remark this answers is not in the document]
arun3528: ya fix it part of that ticket

src/media/local-display-track.ts
arun3528: correct created a new class for LocalComputer audio now

src/media/track.ts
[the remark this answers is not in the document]
arun3528: yep

src/device/device-management.ts
```diff
@@ -1,66 +1,66 @@
...
-   * Creates new error.
-   *
-   * @param type - Error type.
-   * @param message - Error message.
-   */
-  constructor(type: ErrorTypes, message = '') {
-    this.type = type;
-    this.message = message;
+export async function createMicrophoneAndCameraTracks(
+  audioConstraints?: MicrophoneConstraints,
```
bbaldino, line 24: I just noticed: should these support `boolean`s as well? i.e.

```
audioConstraints?: MicrophoneConstraints | boolean
```

src/device/device-management.ts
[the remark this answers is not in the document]
bbaldino: what do you mean 'moved it?

src/error/WebrtcError.ts
```diff
@@ -0,0 +1,25 @@
+export enum ErrorTypes {
+  DEVICE_PERMISSION_DENIED = 'DEVICE_PERMISSION_DENIED',
+  CREATE_CAMERA_TRACK_FAILED = 'CREATE_CAMERA_TRACK_FAILED',
+  CREATE_MICROPHONE_TRACK_FAILED = 'CREATE_MICROPHONE_TRACK_FAILED',
+}
+
+/**
+ * Represents a webrtc core error, which contains error type and error message.
+ */
+export class WebrtcError {
```
bbaldino, line 10: should this derive from `Error`?

src/index.ts
```diff
@@ -7,6 +7,7 @@ export * from './media/local-camera-track';
 export * from './media/local-display-track';
 export * from './media/local-microphone-track';
 export * from './media/local-track';
+export * from './media/local-video-track';
```
bbaldino, line 10: LocalComputerAudioTrack needs to be exported here too I think

src/media/local-track.ts
bbaldino: I don't think I understand.  Nothing that isn't a `LocalTrack` should be firing `UnderlyingTrackChange`, right? So what's the issue with moving the `UnderlyingTrackChange` event to `LocalTrack`?  Something with effects? Do we have support for effects on remote tracks?

src/media/local-track.ts
```diff
@@ -52,77 +34,52 @@ export type TrackEffect = any;
...
    */
-  get muted(): boolean {
-    return !this.underlyingTrack.enabled;
+  disposeEffects(): void {
+    if (this.effects.size > 0) {
+      this.effects.forEach((effect: TrackEffect) => effect.dispose());
+      this.effects.clear();
+
+      // this.underlyingStream = this.originalStream;
+      // this.emit(LocalTrackEvents.UnderlyingTrackChange);
```
bbaldino, line 81: Are these needed? Should either restore them or delete the lines

src/media/local-video-track.ts
bbaldino: I meant that I'm not sure the repetition is that bad, thinking it might be better to just do:
```
type CameraConstraints = Pick<
  MediaTrackConstraints,
  'aspectRatio' | 'height' | 'width' | 'frameRate' | 'deviceId' | 'facingMode'
>;

type DisplayConstraints = Pick<
  MediaTrackConstraints,
  'aspectRatio' | 'height' | 'width' | 'frameRate' | 'deviceId' | 'displaySurface' | 'logicalSurface'
>;
```
(note: I don't see any reference to 'optimizationMode' in the spec?)

src/media/local-video-track.ts
```diff
@@ -0,0 +1,75 @@
...
+
+export interface DisplayConstraints extends BaseDisplayConstraints {
+  displaySurface?: DisplaySurface;
+  logicalSurface?: boolean;
+}
+
+/**
+ * A class to map resolution with video constraints.
+ */
+export const StaticVideoEncoderConstraints: { [key: string]: CameraConstraints } = {
```
bbaldino, line 42: I think the use of 'encoder' here feels a bit inconsistent now with the other types.  I'd just call this something like `PresetCameraConstraints`?

src/media/track.ts
bbaldino: this can be removed now, right?

src/device/device-management.ts
arun3528: done , can do it for video also , it will look little odd as there is already a function for micropone and camera

src/error/WebrtcError.ts
arun3528: done

src/index.ts
arun3528: done

src/media/local-video-track.ts
arun3528: done with the name change 

for the pick scenario the value add we are providing by adding enums like

```
export enum DisplaySurface {
  browser = 'browser',
  monitor = 'monitor',
  window = 'window',
}
```

is the developer know what enum to use , currently its just string which comes out of MediaTrackConstraints

src/media/local-track.ts
arun3528: removed the comment 

i still need to attach events for the effect track for onended and onmuted 

when applying effects i have to save  both the local track and track with effect in one place , because both are in track class the event for track change would first from there 

i dont see any harm firing the event form track class as localTrack also extends the track if it want it can also fire change event

arun3528:
The boolean is removed

PR #14, opened 2022-11-30
chore:  created custom track for audio and video

arun3528:
1) Splits the local-track into local-audio-track and local-video-track
2) remove the logic of saving streams as its not needed for effects , we could create a dynamic streams when starting effects 
3) keeping the tracks basics so that we can use it for remote tracks

NOTE: remote track will come later

### DIFF
--- a/src/media/local-audio-track.ts
+++ b/src/media/local-audio-track.ts
@@ -1,0 +1,40 @@
+import { LocalTrack } from './local-track';
+
+export type AudioDeviceConstraints = {
+  deviceId?: string;
+};
+
+export type StaticAudioEncoderConfig = {
+  low: '16kHz';
+};
+
+export type AudioEncoderConfig = {
+  bitrate?: number;
+  sampleRate?: number;
+  sampleSize?: number;
+  channelCount?: number;
+};
+
+export type MicrophoneConstraints = {
+  echoCancellation?: boolean;
+  autoGainControl?: boolean;
+  noiseSuppression?: boolean;
+  microphoneDeviceId?: string;
+  encoderConfig?: AudioEncoderConfig;
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type TrackEffect = any;
+/**
+ * Represents a audio track for a audio source.
+ */
+export class LocalAudioTrack extends LocalTrack {
+  /**
+   * Makes sure to apply the encoderConfig for the audio.
+   *
+   * @param encoderConfig - Encoder config for audio.
+   */
+  setEncoderConfig(encoderConfig: AudioEncoderConfig): void {
+    this.getMediaStreamTrack().applyConstraints(encoderConfig);
+  }
+}
--- a/src/media/local-microphone-track.ts
+++ b/src/media/local-microphone-track.ts
@@ -1,6 +1,6 @@
-import { LocalTrack } from './local-track';
+import { LocalAudioTrack } from './local-audio-track';
 
 /**
  * Represents a local track for a microphone source.
  */
-export class LocalMicrophoneTrack extends LocalTrack {}
+export class LocalMicrophoneTrack extends LocalAudioTrack {}
--- a/src/media/local-track.ts
+++ b/src/media/local-track.ts
@@ -1,23 +1,8 @@
 /* eslint-disable no-underscore-dangle */
-import { EventEmitter, EventMap } from '../event-emitter';
+import { EventMap } from '../event-emitter';
 import { MediaStreamTrackKind } from '../peer-connection';
 import { logger } from '../util/logger';
-
-export enum LocalTrackEvents {
-  Ended = 'ended',
-  Muted = 'muted',
-  /**
-   * Fires when the published state of a LocalTrack changes.  A value of 'true' indicates the
-   * track has been published, and 'false' indicates that it is no longer published.  Tracks are
-   * unpublished by default, and must be published explicitly (no event will be fired to indicate
-   * the initial state of 'unpublished').
-   */
-  PublishedStateUpdate = 'published-state-update',
-  /**
-   * Fires when there has been a change in the underlying track.
-   */
-  UnderlyingTrackChange = 'underlying-track-change',
-}
+import { Events, Track } from './track';
 
 export interface TrackState {
   id: string;
@@ -39,11 +24,8 @@ export type TrackPublishEvent = {
   trackState: TrackState;
 };
 
-export interface TrackEvents extends EventMap {
-  [LocalTrackEvents.Ended]: (event: TrackEndEvent) => void;
-  [LocalTrackEvents.Muted]: (event: TrackMuteEvent) => void;
-  [LocalTrackEvents.PublishedStateUpdate]: (event: TrackPublishEvent) => void;
-  [LocalTrackEvents.UnderlyingTrackChange]: () => void;
+export interface LocalTrackEvents extends EventMap {
+  [Events.PublishedStateUpdate]: (event: TrackPublishEvent) => void;
 }
 
 // eslint-disable-next-line @typescript-eslint/no-explicit-any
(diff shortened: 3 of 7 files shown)


src/media/local-audio-track.ts
```diff
@@ -0,0 +1,89 @@
...
+export class LocalAudioTrack extends LocalTrack {
+  private effects: Map<string, TrackEffect> = new Map();
+
+  /**
+   * Get an effect by name.
+   *
+   * @param name - The effect name.
+   * @returns A MicrophoneEffect.
+   */
+  getEffect(name: string): TrackEffect | undefined {
```
marcin-bazyl, line 41: why do we need copies of getEffect() and addEffect() in LocalAudioTrack() and LocalVideoTrack()? They are basically the same code, the only difference is that addEffect for audio gets the audio track from the stream, while for video gets the video track, but this can be taken care of by just having a getUnderlyingTrack() function in the effect as each effect will only make sense for either audio or video so the effect will know which track to return, or another way would be to just call `getTracks()[0]` because the underlying stream will always have just 1 track (audio or video), right?

src/media/local-audio-track.ts
```diff
@@ -0,0 +1,89 @@
...
+    this.effects.forEach((effect: TrackEffect) => effect.dispose());
+  }
+
+  /**
+   * Adds background noise cancellation effect.
+   *
+   * @param name - The name of the effect.
+   * @param effect - The effect to add.
+   */
+  async addBNREffect(name: string, effect: TrackEffect): Promise<void> {
```
marcin-bazyl, line 76: why do we need an effect specific function like this? Each effect should just implement the same TrackEffect interface and then can be used with addEffect(), right? Otherwise you will have to keep adding a new function every time you add an effect and effects are not even in the same repo, right? and it's just duplicated code and you will have to check if the right effect instance is added, etc, It feels like it just creates more code to maintain and test but no added value.

src/media/local-audio-track.ts
```diff
@@ -0,0 +1,89 @@
+import { logger } from '../util/logger';
+import { LocalTrack } from './local-track';
+
+export type AudioDeviceConstraints = {
+  deviceId?: string;
+};
+
+export type StaticAudioEncoderConfig = {
+  low: '16kHz';
```
marcin-bazyl, line 9: we need more than just "low" and this should be done in the same way like staticVideoEncoderConfig - with an object containing populated instances of AudioEncoderConfig, so the type would be:
```
export type StaticAudioEncoderConfig: {[key: string]: AudioEncoderConfig}
```

src/media/local-video-track.ts
```diff
@@ -0,0 +1,130 @@
...
+  frameRate?: number | ConstrainLong;
+  height?: number | ConstrainLong;
+  width?: number | ConstrainLong;
+  aspectRatio?: number;
+};
+
+/**
+ * A class to map resolution with video constraints.
+ */
+export const staticVideoEncoderConfig = {
```
marcin-bazyl, line 33: type name should start with upper case letter and it would be better to specify the type here:
```
export type StaticVideoEncoderConfig: {[key: string]: VideoEncoderConfig}
```

src/media/local-microphone-track.ts
```diff
@@ -1,6 +1,6 @@
-import { LocalTrack } from './local-track';
+import { LocalVideoTrack } from './local-video-track';
 
 /**
  * Represents a local track for a microphone source.
  */
-export class LocalMicrophoneTrack extends LocalTrack {}
+export class LocalMicrophoneTrack extends LocalVideoTrack {}
```
marcin-bazyl, line 6: LocalMicrophoneTrack should extend LocalAudioTrack

src/media/local-track.ts
```diff
@@ -39,92 +24,16 @@ export type TrackPublishEvent = {
   trackState: TrackState;
 };
 
-export interface TrackEvents extends EventMap {
-  [LocalTrackEvents.Ended]: (event: TrackEndEvent) => void;
-  [LocalTrackEvents.Muted]: (event: TrackMuteEvent) => void;
-  [LocalTrackEvents.PublishedStateUpdate]: (event: TrackPublishEvent) => void;
-  [LocalTrackEvents.UnderlyingTrackChange]: () => void;
+export interface LocalTrackEvents extends EventMap {
+  [Events.PublishedStateUpdate]: (event: TrackPublishEvent) => void;
```
marcin-bazyl, line 28: where are `Events` defined? they seem to be missing

src/media/local-audio-track.ts
```diff
@@ -0,0 +1,89 @@
...
+  channelCount?: number;
+};
+
+export type MicrophoneConstraints = {
+  echoCancellation?: boolean;
+  autoGainControl?: boolean;
+  noiseSuppression?: boolean;
+  microphoneDeviceId?: string;
+  encoderConfig?: AudioEncoderConfig;
+};
```
bbaldino, line 25: Are there built-in types these could leverage?

src/media/local-audio-track.ts
bbaldino: the `effects` map looks like it could be `protected` in `LocalTrack` as well, and certainly `getEffect` could be there.

src/media/local-audio-track.ts
```diff
@@ -0,0 +1,89 @@
...
+   */
+  async addEffect(name: string, effect: TrackEffect): Promise<void> {
+    await effect.load(new MediaStream([this.getMediaStreamTrack()]));
+    this.setMediaStreamTrackWithEffects(effect.getUnderlyingStream().getAudioTracks()[0]);
+    this.effects.set(name, effect);
+  }
+
+  /**
+   * Cleanup local microphone track.
+   */
```
bbaldino, line 65: This doc seems incorrect

src/media/local-audio-track.ts
```diff
@@ -0,0 +1,89 @@
...
+  }
+
+  /**
+   * Makes sure to apply the encoderConfig for the audio.
+   *
+   * @param encoderConfig - Encoder config for audio.
+   */
+  setEncoderConfig(encoderConfig: AudioEncoderConfig): void {
+    this.getMediaStreamTrack().applyConstraints(encoderConfig);
+  }
```
bbaldino, line 88: what API is WCME currently using to accomplish this?

src/media/local-video-track.ts
```diff
@@ -0,0 +1,130 @@
+import { logger } from '../util/logger';
+import { LocalTrack } from './local-track';
+
+export enum FacingMode {
+  user = 'user',
+  environment = 'environment',
+}
```
bbaldino, line 7: There are built-in types for all these, no?  Can we leverage those?

src/media/local-audio-track.ts
arun3528: ya we could move it there but the main reason i moved it to save the effects in respective class files but does not matter can change

src/media/local-audio-track.ts
arun3528: The only reason we added this one was avoid the name it just encloses  the name , ok

src/media/local-audio-track.ts
arun3528: Not sure how i missed this but @hmcgowan  is working on it

src/media/local-audio-track.ts
arun3528: done

src/media/local-audio-track.ts
arun3528: removed

src/media/local-microphone-track.ts
arun3528: done

src/media/local-audio-track.ts
arun3528: we could but the naming is friendly and helps with documentation  , this is more cleaner for developer to understand they look similar to built ins

src/media/local-audio-track.ts
```diff
@@ -0,0 +1,89 @@
...
+export type MicrophoneConstraints = {
+  echoCancellation?: boolean;
+  autoGainControl?: boolean;
+  noiseSuppression?: boolean;
+  microphoneDeviceId?: string;
+  encoderConfig?: AudioEncoderConfig;
+};
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type TrackEffect = any;
```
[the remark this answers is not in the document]
arun3528: types script was throwing an error if we dont define the types 

currently we are not importing the effects lib as its needs npm token , so waiting for them to export types separately
moved it to local-track

src/media/local-audio-track.ts
arun3528: I have node seen the code which does mid call apply constraints , i can only see createMicrophoneTrack 

and it uses the core as well for that

src/media/local-video-track.ts
arun3528: currently only MediaTrackCapabilities is exposed and not the sub types 
Screenshot 2022-12-14 at 11 41 26 AM

arun3528:
converting this to one PR by fixing the review comments

src/media/local-audio-track.ts
arun3528: done

src/media/local-audio-track.ts
arun3528: done

PR #15, opened 2022-11-30
chore: adding sample app for the media core

arun3528:
This PR just adds samples app to the webrtc core for testing all the features like custom constraints for track creation etc


marcin-bazyl:
Can we maybe use some framework like React? Plain vanilla javascript becomes a pain to maintain very quickly as the app grows with functionality. We have a perfect example of that in the JS-SDK meetings plugin sample app. If we don't start with a framework now at the beginning, it will be even more effort to move to it later. The WCME Homer test app is using react, see here: https://sqbu-github.cisco.com/CPaaS/homer-webrtc-test-app Can we copy that approach?

arun3528:
@marcin-bazyl  we can do that , is it ok if we take that job after this , it required pipeline and webpack setup

marcin-bazyl:
@arun3528 ok, yeah, as long as it's done now as part of the https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-375070 epic and not put off for "later" that never happens. Can you add the jira to the epic?

haohancom:
@marcin-bazyl  I've added: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-389258

arun3528:
closing this as @haohancom  has already done this on this mono PR

PR #17, opened 2022-12-08
Sample

haohancom:
This PR just provide a samples app base on React to the webrtc core, not includes function implementations. Will invoke functions in `device-management.ts` once they are merged.


samples/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "samples",
+  "version": "0.1.0",
+  "private": true,
+  "dependencies": {
+    "@testing-library/jest-dom": "^5.16.5",
```
arun3528, line 6: i am not sure if you need a package.json here , we will need to merger the code together or make them as monorepo

samples/package.json
```diff
@@ -0,0 +1,41 @@
+{
```
arun3528, line 1: also we dont need package.lock file here we are using yarn

samples/package.json
```diff
@@ -0,0 +1,41 @@
...
+    "build": "react-scripts build",
+    "test": "react-scripts test",
+    "eject": "react-scripts eject"
+  },
+  "eslintConfig": {
+    "extends": [
+      "react-app",
+      "react-app/jest"
+    ]
+  },
```
arun3528, line 28: all these need to be converted to configs are moved to the main package.json

Where did you refer the samples from ?

arun3528 (changes requested):

samples/package.json
haohancom: I use [create-react-app](https://create-react-app.dev/) to build this sample folder from zero. You are right, Arun, merge this `package.json` with the main one is better. Let me try it.

PR #18, opened 2022-12-14
feat: samples + mono repo + track changes

haohancom:


haohancom:
@bbaldino  sorry for late response, i missed your comment. it should be `packages` instead of `packets` , I've renamed them.

bbaldino (commented):
I think this looks fine.  But I know @arun3528  has some big changes pending...Arun do you want to get your stuff in first?

arun3528:
Is there a way to fix the issue with file renames they are showing up as new files

haohancom:
> Is there a way to fix the issue with file renames they are showing up as new files

I'm not quite sure what you're referring to... @arun3528 could you say a little more detailed?

arun3528:
Will take that comment back , I thought the branch already had local video
and audio track file and we are just changing it from js to to file

PR #20, opened 2022-12-19
feat: add the api for fetching the connection type

x-epoch:
API for fetching the the connection type, which would return the type of Promise<'UDP' | 'TCP' | 'TURN-TLS' | 'TURN-TCP' | 'TURN-UDP' | 'unknown'>. 
Details can refer to https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-379327

### DIFF
--- a/src/peer-connection.spec.ts
+++ b/src/peer-connection.spec.ts
@@ -10,7 +10,124 @@ jest.mock('./connection-state-handler');
 
 const mockCreateRTCPeerConnection = mocked(createRTCPeerConnection, true);
 
+// eslint-disable-next-line jsdoc/require-jsdoc
+function constructCandidatePairStats(id: string, localCandidateId: string, state: string) {
+  return {
+    id: `${id}`,
+    timestamp: 1671091266890.878,
+    type: 'candidate-pair',
+    transportId: 'T11',
+    localCandidateId: `${localCandidateId}`,
+    remoteCandidateId: 'I+UUFv24B',
+    state: `${state}`,
+  };
+}
+
+// eslint-disable-next-line jsdoc/require-jsdoc
+function constructLocalCandidateStats(
+  id: string,
+  protocol: string,
+  relayProtocol: string | undefined
+) {
+  return {
+    id,
+    timestamp: 1671091266890.878,
+    type: 'local-candidate',
+    transportId: 'T21',
+    isRemote: false,
+    networkType: 'vpn',
+    ip: '2001:420:c0c8:1005::456',
+    address: '2001:420:c0c8:1005::456',
+    port: 53906,
+    protocol,
+    candidateType: 'host',
+    priority: 2122197247,
+    relayProtocol,
+  };
+}
+
 describe('PeerConnection', () => {
+  describe('getCurrentConnectionType', () => {
+    beforeEach(() => {
+      jest.clearAllMocks();
+    });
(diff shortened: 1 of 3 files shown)


src/peer-connection.spec.ts
```diff
@@ -10,7 +10,124 @@ jest.mock('./connection-state-handler');
...
+    port: 53906,
+    protocol,
+    candidateType: 'host',
+    priority: 2122197247,
+    relayProtocol,
+  };
+}
+
 describe('PeerConnection', () => {
+  describe('getCurrentConnectionType', () => {
```
marcin-bazyl, line 50: I think it would be worth adding tests for these cases:
- when no candidate pair has succeeded
- when ice state is not connected/completed

src/peer-connection.ts
```diff
@@ -285,6 +285,45 @@ class PeerConnection extends EventEmitter<PeerConnectionEventHandlers> {
...
+
+  /**
+   * Returns the type of a connection that has been established.
+   *
+   * @returns The connection type which would be {Promise<'UDP' | 'TCP' | 'TURN-TLS' | 'TURN-TCP' | 'TURN-UDP' | 'unknown'>}.
+   */
+  async getCurrentConnectionType(): Promise<string> {
+    // make sure this method only can be called when the ice connection is established;
+    const isIceConnected =
+      this.pc.iceConnectionState === 'connected' || this.pc.iceConnectionState === 'completed';
```
marcin-bazyl, line 297: Our PeerConnection class has a `getConnectionState()` method which we could use here and then you would only need to check against `ConnectionState.Connected` here. However, the `getConnectionState()` method returns the overall state (so it only says "connected" if both ICE and DTLS are connected) while here I think we only care about ICE state, so I don't mind that much which way we do it here, just writing the comment to make you aware that you have the option to use `PeerConnection.getConnectionState()` if you want to

src/peer-connection.ts
```diff
@@ -285,6 +285,45 @@ class PeerConnection extends EventEmitter<PeerConnectionEventHandlers> {
   get iceGatheringState(): RTCIceGathererState {
     return this.pc.iceGatheringState;
   }
+
+  /**
+   * Returns the type of a connection that has been established.
+   *
+   * @returns The connection type which would be {Promise<'UDP' | 'TCP' | 'TURN-TLS' | 'TURN-TCP' | 'TURN-UDP' | 'unknown'>}.
+   */
+  async getCurrentConnectionType(): Promise<string> {
```
bbaldino, line 294: Wondering if it's worth explicitly defining a type for these: instead of `string`, doing something like:

```
const ConnectionType = 'UDP' | 'TCP' | 'TURN-TLS' | 'TURN-TCP' | 'TURN-UDP' | 'unknown';
```

src/peer-connection.spec.ts
x-epoch: Both cases were added already.

src/peer-connection.ts
x-epoch: Thank for the reminder, yeah I see this API in `ConnectionStateHandler`, but as you said this `ConnectionState.Connected` states also represent the DTLS connection state which not actually meet our expectation, so we can just keep the code currently.

src/peer-connection.ts
x-epoch: Good suggestion, I have already updated this code .

marcin-bazyl (approved):

PR #22, opened 2023-01-13
fix: video resolution issue caused by getUserMedia API invoking behavior

x-epoch:
Code change for documenting the behavior that previous video track not stopped or createCameraTrack not return yet which can refer to https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-391260

### DIFF
--- a/src/device/device-management.ts
+++ b/src/device/device-management.ts
@@ -3,6 +3,9 @@ import { LocalCameraTrack } from '../media/local-camera-track';
 import { LocalDisplayTrack } from '../media/local-display-track';
 import { LocalMicrophoneTrack } from '../media/local-microphone-track';
 
+let isCameraTaskInProgress = false;
+let capturedVideoStream: MediaStream;
+
 export enum ErrorTypes {
   DEVICE_PERMISSION_DENIED = 'DEVICE_PERMISSION_DENIED',
   CREATE_CAMERA_TRACK_FAILED = 'CREATE_CAMERA_TRACK_FAILED',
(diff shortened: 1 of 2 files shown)


src/device/device-management.ts
```diff
@@ -51,16 +54,30 @@ export type VideoDeviceConstraints = {
...
 ): Promise<LocalCameraTrack> {
-  let stream: MediaStream;
+  // let stream: MediaStream;
+  if (isCameraTaskInProgress) {
+    throw new WcmeError(
+      ErrorTypes.CREATE_CAMERA_TRACK_FAILED,
+      `camera track can NOT be created since previous call has not been finished yet`
+    );
+  }
+  if (capturedVideoStream && capturedVideoStream.active) {
```
marcin-bazyl, line 64: I don't think we can do that - this will throw an error if the app calls createCameraTrack() more than once in order to have multiple video tracks at the same time

src/device/device-management.ts
```diff
@@ -51,16 +54,30 @@ export type VideoDeviceConstraints = {
...
+    throw new WcmeError(
+      ErrorTypes.CREATE_CAMERA_TRACK_FAILED,
+      `camera track can NOT be created since previous captured video stream does not be stopped`
+    );
+  }
+  isCameraTaskInProgress = true;
```
marcin-bazyl, line 70: this solution won't work if the app calls media.getUserMedia directly

src/device/device-management.ts
x-epoch: this is a unsafe behavior if you call the createCameraTrack() more than once before stopping the first one, 
because the constraint params in second getUserMedia call would not take effect which can refer to https://codepen.io/brownzw/pen/KKeEJPx .

src/device/device-management.ts
x-epoch: Yeah, this solution can not solve this.

PR #23, opened 2023-01-13
docs: documenting the unsafe behaviors for calling createCameraTrack

x-epoch:


src/device/device-management.ts
```diff
@@ -43,7 +43,10 @@ export type VideoDeviceConstraints = {
 };
 
 /**
- * Creates a camera video track.
+ * Creates a camera video track. Please note that the constraint params in second getUserMedia call would NOT take effect when:
+ *
+ * 1. Previous captured video track is not stopped.
```
marcin-bazyl, line 48: should this say "previous captured video track from the same device"?

x-epoch: done

marcin-bazyl (approved):

PR #26, opened 2023-01-31
chore: Import media sdk from bundle

haohancom:
When I ran `yarn start` I found these three functions can not be found, think they are discarded due to rollup tree-shaking. In opposite, `bundle` includes all exported functions. It won't throw function not found error if we use `bundle`.
image

### DIFF
--- a/samples/src/App.js
+++ b/samples/src/App.js
@@ -11,7 +11,7 @@ import {
   getSpeakers,
   getMicrophones,
   getCameras,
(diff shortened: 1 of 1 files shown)


arun3528 (approved):
For now you can add it but we have to move to using the workspace feature rather then the bundle

haohancom:
@arun3528 ya you are right, I'll do investigate of it.

PR #27, opened 2023-02-01
feat: support applying music mode constraints

brycetham:
This PR adds the necessary constraints and methods to support music mode on local audio tracks.

It adds the following APIs to LocalTrack:

- `applyConstraints`: useful for applying music mode constraints to the track, though [it doesn't work for audio constraints on Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=796964); still, I think it'll be useful for other browsers like Firefox, so that it doesn't need to make a new call to `getUserMedia`.
- `getConstraints`: not strictly necessary for music mode but useful for testing purposes.
- `getSettings`: as far as I know, this is the recommended way to get the device ID of the track, useful for calling `getUserMedia` again on Chrome.

### DIFF
--- a/src/device/device-management.ts
+++ b/src/device/device-management.ts
@@ -31,6 +31,9 @@ export class WcmeError {
 
 export type AudioDeviceConstraints = {
   deviceId?: string;
+  autoGainControl?: boolean;
+  echoCancellation?: boolean;
+  noiseSuppression?: boolean;
 };
 
 export type VideoDeviceConstraints = {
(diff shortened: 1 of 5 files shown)


arun3528:
thanks for testing it out, The code base will change significantly with the beta version https://github.com/webex/webrtc-core/tree/beta

just FYI

bbaldino:
lgtm.  @arun3528  should he do this against the beta branch instead?

brycetham:
@bbaldino I think we're going to wait until the beta branch is merged in, and then merge this PR in after.

bbaldino (approved):

PR #28, opened 2023-02-07
chore: update the samples 

arun3528:
updated the samples to work with the existing webrtc core 
import effects and make it work for blur
fixed the samples app build issue


samples/src/App.js
```diff
@@ -32,41 +32,142 @@ function App() {
...
+
+  const enableVirtualBackground = (enabled) => {
+
+    if(!videoEffect) {
+      effect = new VirtualBackgroundEffect({
+        mode: "IMAGE",
+        bgVideoUrl: VIDEO_URL,
+      });
+
+      localCameraTrack.addEffect("virtualBackground", effect)
```
bbaldino, line 65: should there be a call to `setVideoEffect` in here?

samples/src/App.js
```diff
@@ -32,41 +32,142 @@ function App() {
...
+  const enableBlurBackground = async (enabled) => {
+    let effect = null;
+    if(!blurEffect) {
+      effect = new VirtualBackgroundEffect({
+        mode: `BLUR`,
+        blurStrength: `STRONG`,
+        quality: `HIGH`,
+      });
+      await localCameraTrack.addEffect("blur", effect);
+      setBlurEffect(effect)
```
bbaldino, line 86: does it make sense to create this even if enabled is false?

bbaldino: if so, maybe it should just be created by default instead of in the first call to `enableBlurBackground`?

samples/src/App.js
```diff
@@ -32,41 +32,142 @@ function App() {
...
+        }
+          
+
+
+        localDisplayTrack.on('muted',() =>{
+
+          })
+          localDisplayTrack.on('ended',() =>{
+            
+          })
```
bbaldino, line 167: empty? also looks like indentation is wrong

samples/src/components/MediaDevices.jsx
```diff
@@ -70,13 +70,15 @@ export default (props) => {
         </Col>
         <Col>
           <Button
+            variant={audioStatus.muted?"danger" : 'primary'}
```
bbaldino, line 73: muted/unmuted = "danger"/"primary" ?

samples/src/components/MediaDevices.jsx
```diff
@@ -70,13 +70,15 @@ export default (props) => {
...
             onClick={() => {
-              props.muteAudioTrack();
+              setAudioStatus({...audioStatus, muted: !audioStatus.muted})
+              props.muteAudioTrack(audioStatus.muted);
             }}
           >
-            Mute
+            {audioStatus.muted?"unMute" : 'Mute'}
```
bbaldino, line 79: spacing, inconsistent quotes

samples/src/App.js
arun3528: i can add it back , previously i remove it based on the review comments from either you or marcin dont remember

samples/src/components/MediaDevices.jsx
haohancom: @bbaldino in bootstrap, different `variant`s are mapped with different specified css format. In brief, `danger` and `primary` here mean different button color.

samples/src/App.js
arun3528: its up to the developer does not make a difference , creating it ahead or later

samples/src/App.js
arun3528: yep need to fill it up, thought will do it later , will do it

samples/src/components/MediaDevices.jsx
arun3528: ha ha just for UI button to be red or blue

samples/src/components/MediaDevices.jsx
arun3528: i need to actually add jsx eslint in the samples , have as a todo

samples/src/App.js
arun3528: @bbaldino  my bad we cannot change it directly as addEffect is part of the local-track , I could add it on local-video-track  but not much difference

samples/src/App.js
bbaldino: Right, but the point of the sample is to show the idiomatic way of doing it, so I think we should make sure the example is clear.

samples/src/components/MediaDevices.jsx
bbaldino: gotcha :)

samples/src/App.js
arun3528: fixed this

samples/src/App.js
arun3528: with react it gets confusing , thought it would be better to keep the effect creation and management in a single place 

we are not going to use this demo for developers , There will be a separate doc on the SDK side for the usage 

i can move the VBG creation on the useState when the app loads if you want

bbaldino (approved):

PR #32, opened 2023-02-21
feat: new track management

arun3528:
The beta version which is 1.4.0 does have the new way of track management with effects integrated as well


packages/webrtc-core/src/media/track.ts
```diff
@@ -0,0 +1,199 @@
...
+
+/**
+ * Basic Track class.
+ */
+export abstract class Track extends EventEmitter<TrackEvents> {
+  static Events = Events;
+
+  ID: string;
+
+  kind: string;
```
brycetham, line 70: Should this be type `TrackKind`?

packages/webrtc-core/src/media/track.ts
```diff
@@ -0,0 +1,199 @@
...
+    this.#mediaStreamTrack.onended = () => {
+      this.emit(Events.Ended, { trackState: this.trackState });
+    };
+
+    /**
+     * Emit mute event when the underlying track gets muted.
+     */
+    this.#mediaStreamTrack.onmute = () => {
+      this.emit(Events.Muted, {
+        trackState: {
```
brycetham, line 106: This is incompatible with `TrackState` in `LocalTrack`. Perhaps `LocalTrack` should override this event?

packages/webrtc-core/src/media/track.ts
```diff
@@ -0,0 +1,199 @@
...
+   * @param stream - The MediaStream for this LocalTrack.
+   * @param track
+   */
+  constructor(track: MediaStreamTrack) {
+    super();
+    this.ID = track.id;
+    this.status = track.readyState as TrackStatus;
+    this.label = track.label;
+    this.kind = track.kind;
+    this.#mediaStreamTrack = track;
```
brycetham, line 92: Should we also assign `mediaStreamTrackWithEffect` to `track` here and make it not optional? In the original code, `LocalTrack` assigns both `originalStream` and `underlyingStream` in the constructor.

packages/webrtc-core/src/media/track.ts
arun3528: fixing it

packages/webrtc-core/src/media/track.ts
arun3528: we do have  #mediaStreamTrackWithEffect?: MediaStreamTrack; in the track but not optional as a developer might not need effects in some cases so that would be null

packages/webrtc-core/src/index.ts
```diff
@@ -2,11 +2,15 @@
...
 
+// TODO: webex js sdk re export the types for device-manager
+
 export * from './device/device-management';
 export * from './media/local-camera-track';
 export * from './media/local-display-track';
+export * from './media/local-computer-audio-track';
 export * from './media/local-microphone-track';
 export * from './media/local-track';
+export * from './media/local-video-track';
```
brycetham, line 13: I think you're missing `export * from './media/local-audio-track';` from here as well.

packages/webrtc-core/src/media/track.ts
```diff
@@ -0,0 +1,198 @@
...
+    }
+  }
+
+  /**
+   *
+   * This method returns the actual MediaStreamTrack.
+   *
+   * @returns #mediaStreamTrack of type MediaStreamTrack.
+   */
+  getMediaStreamTrack(): MediaStreamTrack {
```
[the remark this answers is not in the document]
arun3528: i spoke with andi about this the last time before removing this logic . The effect lib is using steams but there is always one track in there . It does not makes send to save streams then just keep swapping the tracks in the webrtc core . As the stream is only for the effects
cc @andiwils

packages/webrtc-core/src/media/track.ts
arun3528: @bbaldino  do you have a better name for this ?

packages/webrtc-core/src/index.ts
arun3528: will check and add it

arun3528: fixed it

packages/webrtc-core/src/media/track.ts
arun3528: cc @marcin-bazyl

marcin-bazyl: I agree with Bryce it would be nice to be able to just call a single method and always get a track (with effect if effects are applied, without it if there are no effects). I think we can keep existing name `getMediaStreamTrackWithEffects()` and just return the `track` if there are no effects.

arun3528: ok

arun3528: @brycetham  will change it

packages/webrtc-core/src/media/track.ts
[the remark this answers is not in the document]
arun3528: @bbaldino do you thinking having a play function makes sense in this case 

we could create a steam for them and play it , i dont want to introduce a getMediaStream function in a track class

packages/webrtc-core/src/media/track.ts
```diff
@@ -171,7 +171,7 @@ export abstract class Track extends EventEmitter<TrackEvents> {
    * @returns #mediaStreamTrack of type MediaStreamTrack.
    */
   getMediaStreamTrackWithEffects(): MediaStreamTrack | undefined {
```
brycetham, line 173: Since `#mediaStreamTrack` is required, we can remove `undefined` from the signature now.

arun3528: will remove it in the next PR waiting for brian comment

.github/workflows/npm-publish.yml
```diff
@@ -7,20 +7,23 @@ on:
...
       - main
+      - beta
+      - alpha
 
 jobs:
   semantic-release:
     name: semantic-latest
     runs-on: ubuntu-latest
+    env:
+      CONNECT_TOKEN: ${{ secrets.CONNECT_TOKEN }}
```
marcin-bazyl, line 18: why do we need this?

packages/webrtc-core/.releaserc
```diff
@@ -1,6 +1,18 @@
...
+    {
+      "name": "beta",
+      "prerelease": true
+    },
+    {
+      "name": "alpha",
+      "prerelease": true
+    },
+    {
+      "name": "mono",
```
marcin-bazyl, line 13: is this needed?

packages/webrtc-core/src/media/track.ts
```diff
@@ -0,0 +1,198 @@
...
+  LIVE = 'live',
+}
+
+// eslint-disable-next-line no-shadow
+export enum TrackKind {
+  AUDIO = 'audio',
+  VIDEO = 'video',
+}
+
+export type TrackPublishEvent = {
```
marcin-bazyl, line 48: this is already defined in local-track.ts

packages/webrtc-core/src/media/track.ts
```diff
@@ -0,0 +1,198 @@
...
+export enum Events {
+  Ended = 'ended',
+  Muted = 'muted',
+  /**
+   * Fires when the published state of a LocalTrack changes.  A value of 'true' indicates the
+   * track has been published, and 'false' indicates that it is no longer published.  Tracks are
+   * unpublished by default, and must be published explicitly (no event will be fired to indicate
+   * the initial state of 'unpublished').
+   */
+  PublishedStateUpdate = 'published-state-update',
```
marcin-bazyl, line 16: does this belong here? only LocalTrack has setPublished() method, not the Track

packages/webrtc-core/src/media/track.ts
```diff
@@ -0,0 +1,198 @@
...
+  #mediaStreamTrackWithEffect?: MediaStreamTrack;
+
+  /**
+   * Constructor for the Track class. Creates an empty CoreLocalTrack or uses an existing one.
+   *
+   * @param track - Media stream track.
+   */
+  constructor(track: MediaStreamTrack) {
+    super();
+    this.ID = track.id;
```
marcin-bazyl, line 88: do we need this if we already have a getter for id that returns the underlying track's id?
and kind of a similar question for label and kind - maybe we should just have getters for them? is there some benefit of having the copies?

packages/webrtc-core/src/media/track.ts
```diff
@@ -0,0 +1,198 @@
...
+    this.#mediaStreamTrack.onended = () => {
+      this.emit(Events.Ended, { trackState: this.trackState });
+    };
+
+    /**
+     * Emit mute event when the underlying track gets muted.
+     */
+    this.#mediaStreamTrack.onmute = () => {
+      this.emit(Events.Muted, {
+        trackState: {
```
marcin-bazyl, line 106: in `this.#mediaStreamTrack.onended` we just use `this.trackState`, should we do the same here?

packages/webrtc-core/src/media/track.ts
```diff
@@ -0,0 +1,198 @@
...
+     */
+    this.#mediaStreamTrack.onended = () => {
+      this.emit(Events.Ended, { trackState: this.trackState });
+    };
+
+    /**
+     * Emit mute event when the underlying track gets muted.
+     */
+    this.#mediaStreamTrack.onmute = () => {
+      this.emit(Events.Muted, {
```
marcin-bazyl, line 105: I'm not sure if it's a good idea to be emitting Events.Muted here. We are emitting `Events.Muted` elsewhere when someone mutes the track (so `enabled` property changes) while here we are in the `onmute` handler - that's for `muted` property change - it's a different thing, it means that the source is for some reason not providing the data, so for example after this gets executed user might try to call seMuted(false) to unmute but it won't work, because the track is not muted, it just doesn't have any input data coming in, because for example the camera is broken.  The `muted` property is read-only. On the other hand, I guess if user has a physical button on an external mic/camera to mute it then pressing it would trigger onmute so maybe it's OK to emit `Events.Muted` here.... I don't know.... @bbaldino do you have any thoughts on this? Maybe we should do this only for LocalTrack not Track, because for remote tracks `muted` is true when for example decoder fails to decode the incoming frames for some reason.

packages/webrtc-core/src/media/track.ts
```diff
@@ -0,0 +1,198 @@
...
+    return this.#mediaStreamTrackWithEffect || this.#mediaStreamTrack;
+  }
+
+  /**
+   *
+   * This method sets the track after applying effect.
+   *
+   * @param track - Media stream track.
+   */
+  setMediaStreamTrackWithEffects(track: MediaStreamTrack): void {
```
marcin-bazyl, line 183: this should be a private method

marcin-bazyl (commented):
I didn't look at the sample app or the tests.

packages/webrtc-core/src/index.ts
```diff
@@ -5,8 +5,11 @@ import * as media from './media';
...
+export * from './media/local-audio-track';
 export * from './peer-connection';
 export * from './peer-connection-utils';
 export { media };
```
brycetham, line 15: I think WebrtcError needs to be exported here as well (I think this was WcmeError before, and that was being exported in the original code).

.github/workflows/npm-publish.yml
arun3528: i will upgrade this but wanted this to test the effects

packages/webrtc-core/.releaserc
arun3528: removed

packages/webrtc-core/src/device/device-management.ts
```diff
@@ -0,0 +1,161 @@
...
+import { LocalCameraTrack } from '../media/local-camera-track';
+import { LocalDisplayTrack } from '../media/local-display-track';
+import { LocalMicrophoneTrack } from '../media/local-microphone-track';
+import { CameraConstraints, DisplayConstraints } from '../media/local-video-track';
+
+import { ErrorTypes, WebrtcError } from '../error/WebrtcError';
+import { LocalComputerAudioTrack } from '../media/local-computer-audio-track';
+
+/**
+ * Creates a camera video track. Please note that the constraint params in second getUserMedia call would NOT take effect when:
```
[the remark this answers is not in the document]
arun3528: done

packages/webrtc-core/src/device/device-management.ts
```diff
@@ -0,0 +1,161 @@
...
+): Promise<[LocalMicrophoneTrack, LocalCameraTrack]> {
+  let stream: MediaStream;
+  try {
+    stream = await media.getUserMedia({
+      audio: audioConstraints,
+      video: videoConstraints,
+    });
+  } catch (error) {
+    throw new WebrtcError(
+      ErrorTypes.CREATE_CAMERA_TRACK_FAILED,
```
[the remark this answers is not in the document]
arun3528: changed to CREATE_MICROPHONE_CAMERA_TRACK_FAILED

packages/webrtc-core/src/device/device-management.ts
```diff
@@ -0,0 +1,161 @@
...
+    stream = await media.getUserMedia({ audio: constraints });
+  } catch (error) {
+    throw new WebrtcError(
+      ErrorTypes.CREATE_MICROPHONE_TRACK_FAILED,
+      `Failed to create microphone track ${error}`
+    );
+  }
+  return new LocalMicrophoneTrack(stream.getAudioTracks()[0]);
+}
+export interface createDisplayTrackConstraint {
```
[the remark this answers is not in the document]
arun3528: done changed the name too

packages/webrtc-core/src/device/device-management.ts
```diff
@@ -0,0 +1,161 @@
...
+ * @returns A Promise that resolves to a LocalDisplayTrack.
+ */
+export async function createDisplayTrack({
+  constraints,
+  withAudio,
+}: createDisplayTrackConstraint): Promise<{
+  localDisplayTrack: LocalDisplayTrack;
+  localComputerAudioTrack?: LocalComputerAudioTrack;
+}> {
+  const stream = await media.getDisplayMedia({ video: constraints, audio: withAudio });
```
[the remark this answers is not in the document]
arun3528: i will change it , i like it to be promise based , i remember older api had bugs where it was throwing errors to , hopefully they are fixed

packages/webrtc-core/src/media/local-track.ts
```diff
@@ -0,0 +1,107 @@
+/* eslint-disable no-underscore-dangle */
+import { EventMap } from '../event-emitter';
+import { logger } from '../util/logger';
+import { Events, Track, TrackState } from './track';
+
+export type TrackEndEvent = {
```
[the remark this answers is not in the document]
arun3528: ya cleaned up the exports

packages/webrtc-core/src/media/track.ts
arun3528: done

packages/webrtc-core/src/media/track.ts
arun3528: if we end up having remote track extending track  that should not have publish function so kept it on local track

packages/webrtc-core/src/media/track.ts
arun3528: good question .

we can put all 3 as getter or keep the ID as uuid because it could change on device change or adding effects !?

any thoughts @bbaldino

packages/webrtc-core/src/device/device-management.ts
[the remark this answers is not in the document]
bbaldino: did you push the change? i still see the old error type

packages/webrtc-core/src/device/device-management.ts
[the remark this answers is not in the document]
bbaldino: don't see this change either

packages/webrtc-core/src/media/local-track.ts
```diff
@@ -0,0 +1,107 @@
...
+};
+
+export interface LocalTrackEvents extends EventMap {
+  [Events.PublishedStateUpdate]: (event: TrackPublishEvent) => void;
+}
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+export type TrackEffect = any;
+// TBD: Fix this once types are published separately
+// export type TrackEffect = BaseMicrophoneEffect | BaseCameraEffect;
```
bbaldino, line 26: do we need to have the effects published separately before merging this?

andiwils:
@arun3528 You should be able to integrate the new public [web-media-effects](https://github.com/webex/webrtc-core/pull/32/) library into this PR. You just need to pass an `authToken` into the constructor of each effect and you won't need to use the `.npmrc` anymore or whatever you're doing for an NPM token.

packages/webrtc-core/src/device/device-management.ts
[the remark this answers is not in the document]
arun3528: forgot to merge it

packages/webrtc-core/src/media/track.ts
bbaldino: publish state doesn't make any sense for a remote track?

packages/webrtc-core/src/media/track.ts
bbaldino: yeah i don't think we need extra state if we already have the track and can retrieve the id from there

bbaldino: plus the getter below doesn't even seem to use this field?

packages/webrtc-core/src/media/track.ts
bbaldino: @arun3528  ^^^

packages/webrtc-core/src/media/track.ts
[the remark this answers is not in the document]
bbaldino: we already agreed no 'play' concept in webrtc-core.

packages/webrtc-core/src/media/track.ts
arun3528: Test the device change with and without deviceID mentioned when getting the steams

packages/webrtc-core/src/media/track.ts
arun3528: done

packages/webrtc-core/src/media/track.ts
arun3528: changed it to UUID

packages/webrtc-core/src/media/track.ts
arun3528: cc @bbaldino

packages/webrtc-core/src/media/track.ts
arun3528: changed to protected

packages/webrtc-core/src/index.ts
arun3528: done

packages/webrtc-core/src/media/local-track.ts
arun3528: Its already published seperately but in the webrtc core lib the only need is the types , 

@andiwils  any suggestion on how to just import the types in the webrtc core we dont want to include the code as the developer will import the effects lib and pass to us

packages/webrtc-core/src/media/local-track.ts
```diff
@@ -0,0 +1,107 @@
...
+
+  /**
+   * Disposes the effect already applied .
+   */
+  disposeEffects(): void {
+    if (this.effects.size > 0) {
+      this.effects.forEach((effect: TrackEffect) => effect.dispose());
+      this.effects.clear();
+
+      // TODO: on dispose restore the track and trigger trackChange event
```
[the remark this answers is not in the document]
arun3528: will change it in this PR thought of doing it in the next PR

arun3528:
@bbaldino  have fixed the review comments 

the publish is just a enum and does not get used in the track file , we export it with other events

bbaldino:
We'll incorporate the concepts in separate PRs at a later time.

PR #36, opened 2023-03-04
New style of events

bbaldino:
This PR introduces a new style of doing events that will be more event to classes in a hierarchy defining their own events at each layer.

I made many attempts at doing an "inhert-from-event-emitter" style approach, and couldn't get any of them to work consistently, so I gave this a try.  This PR defines a "typed event" and builds subscription and emission into the event itself, rather than the class.  There are a couple tradeoffs here:

Pros:
1. No fancy type-system metaprogramming.  While this does leverage EventEmitter under the covers, it doesn't expose any of the type weirdness to the classes itself, which is nice.
2. Plays nicely with inheritance.  These are just members, so we don't need to worry about getting the inheritance/APIs/access at each level correctly, these act just like normal members.
3. It gives classes the flexibility to decide who should be able to emit an event: everyone (internal and external), internal only (the class and its subclasses) or only the class itself.  This is controlled by the accessibility modifier on the event field (public, protected, or private).

Cons:
1. The syntax is unfamiliar compared to traditional event handler APIs, which usually exist on the class.  For example:

Traditional way: `myClass.on('SomeEvent', () => { ... });`
With this PR: `myClass.someEvent.on(() => { ... });`

2. Discoverability: usually a developer can type `someClass.on(` and get some auto-complete hints as to the available events.  Since there is no common `on` API at the class, this is gone.  I've tried to address this somewhat in this PR by grouping the event members under an `Events` field in the class, so you'd do: `someClass.Events.` and then get autocomplete hints as to all the events evident there.

Other notes:
* I also split up the event interfaces a bit: `TypedEventImpl` defines the "complete" interface (i.e. it defines both subscription management and emission APIs).  `TypedEvent` defines only the subscription management APIs, _not_ emission: the thinking being that external entities typically shouldn't be able to call `emit` for a class.
* There's a problem here with jest that I haven't fixed yet.  Currently we validate some subscription behaviors by validating that 'on' methods were called, but now that the calls are on members, not methods, of a class, there's some funkiness in jest that needs to be addressed.  I think it should be doable...I just got tired of fighting it right now :)

### DIFF
--- a/src/peer-connection-utils.spec.ts
+++ b/src/peer-connection-utils.spec.ts
@@ -20,14 +20,16 @@ describe('getLocalDescriptionWithIceCandidates', () => {
     const promise = new PromiseHelper(
       getLocalDescriptionWithIceCandidates(mockPc as unknown as PeerConnection)
     );
-    expect(mockPc.on.mock.calls).toHaveLength(1);
-    expect(mockPc.on.mock.calls[0][0]).toBe(PeerConnection.Events.IceGatheringStateChange);
+    expect(mockPc.Events.iceGatheringStateChange.on.mock.calls).toHaveLength(1);
+    //expect(mockPc.on.mock.calls).toHaveLength(1);
+    //expect(mockPc.on.mock.calls[0][0]).toBe(PeerConnection.Events.IceGatheringStateChange);
     // Grab the listener that was installed
-    const iceGatheringStateListener = mockPc.on.mock.calls[0][1];
+    //const iceGatheringStateListener = mockPc.on.mock.calls[0][1];
+    const iceGatheringStateListener = mockPc.Events.iceGatheringStateChange.on.mock.calls[0][0];
     iceGatheringStateListener({
       target: {
         iceGatheringState: 'gathering',
-      },
+      } as unknown as EventTarget,
     });
     // The reason for these nested calls to Promise.resolve is because, unlike many tests around
     // Promises, here we're trying to validate that a promise _wasn't_ resolved when it shouldn't
@@ -41,7 +43,7 @@ describe('getLocalDescriptionWithIceCandidates', () => {
       iceGatheringStateListener({
         target: {
           iceGatheringState: 'complete',
-        },
+        } as unknown as EventTarget,
       });
       // Make sure the helper has had a chance to resolve its promise before we check again
       Promise.resolve().then(() => {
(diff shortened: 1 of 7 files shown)


src/peer-connection-utils.spec.ts
```diff
@@ -20,14 +20,16 @@ describe('getLocalDescriptionWithIceCandidates', () => {
...
+    //expect(mockPc.on.mock.calls[0][0]).toBe(PeerConnection.Events.IceGatheringStateChange);
     // Grab the listener that was installed
-    const iceGatheringStateListener = mockPc.on.mock.calls[0][1];
+    //const iceGatheringStateListener = mockPc.on.mock.calls[0][1];
+    const iceGatheringStateListener = mockPc.Events.iceGatheringStateChange.on.mock.calls[0][0];
     iceGatheringStateListener({
       target: {
         iceGatheringState: 'gathering',
-      },
+      } as unknown as EventTarget,
```
bbaldino, line 32: This is actually something I'm surprised ever worked, but I think it's actually kinda better we have to do this.  The argument to this event is `EventTarget` which defines no actual fields: it's basically a form of `any`.  In a real browser, the target here will be the entire `RTCPeerConnection`.  I thought about changing the event to match that, but that might cause problems in the browser and we don't want to have to make a type look like that in the tests, so I did this instead. 

I don't get why this wasn't necessary before, though, as what we passed here definitely _doesn't_ match the definition of `EventTarget`.

arun3528:
Looks good for the peerconnection files as its more internal but , for track i would not change this now as all the developers facing events are more on("event:Name") based one of the reason for that is we can have scoped events like 

"event:rooms:added" | "event:rooms"

all the SDK and developers are using the on , to keep it consistent for developers lets not change it for track 

@marcin-bazyl  @brycetham  any thoughts

bbaldino:
> Looks good for the peerconnection files as its more internal but , for track i would not change this now as all the developers facing events are more on("event:Name") based one of the reason for that is we can have scoped events like
> 
> "event:rooms:added" | "event:rooms"
> 
> all the SDK and developers are using the on , to keep it consistent for developers lets not change it for track
> 

Different event implementations/styles for different classes in the same repo isn't a good approach, I don't think.  The problem with the existing events is they don't allow for defining events at the appropriate level of the class hierarchy, and tracks are what need that ability more than anything.  I don't love having to change the syntax either, so I'm all for an implementation that can maintain it and allow for the inheritance scheme described here, but so far this is all I've been able to come up with.

bbaldino:
As of now we plan to use ts-events which has an API that preserves the existing behavior.

PR #38, opened 2023-03-27
fix: made create*Track() functions generic

marcin-bazyl:
To allow JS-SDK to have classes that extend webrtc-core LocalMicrophoneTrack, LocalCameraTrack etc.

### DIFF
--- a/src/device/device-management.spec.ts
+++ b/src/device/device-management.spec.ts
@@ -26,7 +26,7 @@ describe('Device Management', () => {
     it('should call getUserMedia', async () => {
       expect.assertions(1);
 
-      await createMicrophoneTrack({ deviceId: 'test-device-id' });
+      await createMicrophoneTrack(LocalMicrophoneTrack, { deviceId: 'test-device-id' });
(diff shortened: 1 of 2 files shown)


src/mocks/create-permission-status.ts
```diff
@@ -11,6 +11,7 @@
  * @returns PermissionStatus.
  */
 export const createPermissionStatus = (state: PermissionState): PermissionStatus => ({
+  name: 'mock permission status',
```
marcin-bazyl, line 14: I had to add this to have tests passing locally, not sure why this was missing on master

marcin-bazyl: but then it fails in jenkins with it, so reverted it, something is weird here, but don't have time to investigate it now

bbaldino (approved):
lgtm.  note that i expect these, to some extent, to be temporary changes.  i have a writeup on some thoughts of how to improve the design in webrtc-core [here](https://jira-eng-gpk2.cisco.com/jira/browse/WEBEX-325043)

PR #39, opened 2023-03-28
feat: add support for videocontenthint on display tracks

bbaldino:


src/media/local-display-track.ts
```diff
@@ -1,6 +1,33 @@
...
+
+  /**
+   * Create a LocalDisplayTrack from the given values.
+   *
+   * @param stream - The MediaStream for this track.
+   * @param videoContentHint - An optional content hint, describing the content of the track.
+   */
+  constructor(stream: MediaStream, videoContentHint?: VideoContentHint) {
+    super(stream);
+    this._videoContentHint = videoContentHint;
```
marcin-bazyl, line 22: shouldn't we be also setting that content hint on the actual underlying webrtc track?

bbaldino: Ha, yes fair point.  Obviously we're mainly using this setting to signal to the backend, but might as well.

bbaldino: done

brycetham (approved):

PR #43, opened 2023-04-13
feat(bnr): add getEffects and modify disposeEffects

lismith2-cisco:
Need a better interface for applying effects for implementing bnr in webex web client

### DIFF
--- a/src/media/local-track.ts
+++ b/src/media/local-track.ts
@@ -261,11 +261,20 @@ export abstract class LocalTrack extends EventEmitter<TrackEvents> {
   }
 
   /**
+   * Get a copy of the effects Map.
+   *
+   * @returns A copy of the effects map.
+   */
+  getEffects(): Map<string, TrackEffect> {
+    return new Map(this.effects);
+  }
+
+  /**
    * Cleanup the local microphone track.
    */
-  disposeEffects(): void {
+  async disposeEffects(): Promise<void> {
     if (this.effects.size > 0) {
-      this.effects.forEach((effect: TrackEffect) => effect.dispose());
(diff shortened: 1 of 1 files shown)


bbaldino (approved):

bbaldino:
Out of curiosity, what's the use case to get a map of all the effects?

lismith2-cisco:
> Out of curiosity, what's the use case to get a map of all the effects?

 When selecting 'no optimization' I want to disable any current effect (but not dispose it) so it can be re-enabled quickly. There were a couple of options for how to do it, but this seemed best.

bbaldino:
> When selecting 'no optimization' I want to disable any current effect (but not dispose it) so it can be re-enabled quickly. There were a couple of options for how to do it, but this seemed best.

@shaofan-qi I have a vague memory of you finding some issues around disabling but not "removing" effects, am I remembering that right?  Should effects be removed completely when they're not in use?  cc @andiwils who might have some thoughts as well.

bbaldino:
Some comments from @andiwils offline:

> If they will never be used again, they should be disposed of. With VBG, disabling still ate up a bunch of CPU until recently when we decided to just swap the original track with the effect track. We are not doing that with BNR yet but plan to.
> In other words, disabling should have similar performance to not using the effect at all. That much is true with VBG today. Quality shouldn't be affected.

So sounds like even if this might currently have some issues, it's how things _should_ work.

shaofan-qi:
The issue we met was about BNR in Safari. Even the effect was disabled, it resulted in stuttering audio. We may need to look into what's going wrong with the BNR effect. If I remember right, the issue disppears in some newer Safari versions (16.0+) on Intel Macs, but still exists on Apple Silicon Macs, but we haven't tested that for a while.

PR #48, opened 2023-05-04
fix: keep track of loading effects

martroos:
If an effect takes a long time to load (for example, due to a slow network if fetching a resource), and the `addEffect` method is called multiple times with different instances for the same effect name, only one of the effects will actually be applied, and the others will be loaded, but will not be in use, which can consume resources (and potentially cause other issues depending on what the effect does).

This change keeps track of the loading effects, and if, once loading completes, it is determined that the effect is no longer required, the effect will be disposed, and the promise returned from the `addEffect` method will be rejected (to allow consumers to handle this in an appropriate way).

This also clears the loading effects map in the `disposeEffects`, which will also signal that any loading effect is no longer required, since the intention at the time of calling was to reset the effects applied to the track.

### DIFF
--- a/src/media/local-track.spec.ts
+++ b/src/media/local-track.spec.ts
@@ -2,6 +2,22 @@ import { createMockedStream } from '../util/test-utils';
 import { LocalTrack } from './local-track';
 
 /**
+ * Create a mocked track effect.
+ *
+ * @returns A mocked TrackEvent.
+ */
+const createMockedTrackEffect = () => {
+  const effectStream = createMockedStream();
+  const effect = {
+    dispose: jest.fn().mockResolvedValue(undefined),
+    getUnderlyingStream: jest.fn().mockReturnValue(effectStream),
+    load: jest.fn().mockResolvedValue(undefined),
+  };
+
+  return { effectStream, effect };
+};
+
+/**
  * A dummy LocalTrack implementation so we can instantiate it for testing.
  */
 class TestLocalTrack extends LocalTrack {}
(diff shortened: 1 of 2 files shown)


bbaldino:
What's the scenario where multiple calls to load the same effect are made?  Also, if they're loading the same effect, would it not be better to use the first one rather than the last one?

lismith2-cisco:
> What's the scenario where multiple calls to load the same effect are made? Also, if they're loading the same effect, would it not be better to use the first one rather than the last one?

This is a fix for https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-425825 and https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-425826 . I don't directly know the answers to your questions, but Martin won't be in work until Tues, so unlikely to answer before then.

martroos:
> What's the scenario where multiple calls to load the same effect are made? Also, if they're loading the same effect, would it not be better to use the first one rather than the last one?

This fix is actually mostly to handle correctly disposing of an effect when it is no longer needed, when `disposeEffects` is called whilst an effect is loading (hence setting the loading effects to undefined in that method). The bugs @lismith2-cisco provided are the result of effects not being cleared up properly / claiming to be loaded when they shouldn't be due to disposing whilst loading:

* [SPARK-425826](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-425826) - This is a simple case of calling `addEffect` again for the same effect name before the previous effect has loaded - the design of the `LocalTrack` appears to only allow for one effect per name, but only keeps track of effects once loaded, so there is currently a time where an effect is being loaded, but not known to the track.
* [SPARK-425825](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-425825) - This is a result of effectively disposing of a `LocalTrack` whilst an effect is loading on it. This happens when the user chooses to change audio input device (from headset to laptop mic for example), so the original track is no longer needed. The app disposes of the effects on the original track, but since this does not know about loading effects, these will be missed.

Whilst these are not common use cases, it is possible to replace an effect whilst loading - hypothetically it may have some updated parameters - and since loading time can depend on many things, there are no guarantees the load will have completed before an effect is attempted to be replaced. This is also the reason why the last effect is desired instead of the first one - the effects may not be the same, and a general expectation is that the last call to a function takes priority.

marcin-bazyl (approved):

bbaldino (approved):
There are some changes coming to the webrtc-core track classes and I think we'll revisit some of how this works then--maybe we can improve how this looks a bit, but this looks good for now.

bbaldino (approved):
There are some changes coming to the webrtc-core track classes and I think we'll revisit some of how this works then--maybe we can improve how this looks a bit, but this looks good for now.

PR #50, opened 2023-05-09
feat: integrate new stream types

bbaldino:
This PR gets rid of the old track classes and changes things to use the new stream types.


src/device/device-management.ts
```diff
@@ -70,20 +72,18 @@ export async function createCameraTrack<T extends LocalTrack>(
...
 
 /**
- * Creates a microphone audio track.
+ * Creates a LocalMicrophoneStream with the given constraints.
  *
- * @param constructor - Constructor for the local microphone track.
  * @param constraints - Audio device constraints.
  * @returns A LocalTrack object or an error.
  */
-export async function createMicrophoneTrack<T extends LocalTrack>(
```
mccarthytyler, line 83: The `<T extends LocalTrack>` may need to stay in, I think it's used in some cases to create a custom type of stream wrapper.  Maybe `<T extends LocalMicrophoneStream>`?

mccarthytyler (approved):
Looks good, just one question.

src/device/device-management.ts
bbaldino: The plan is to do this differently, where extend of extending these tracks (which then means it's disconnected from the rest of the hierarchy)--they'd instead create and wrap this track, so if all goes well then the `T extends XXX` thing will not be needed.

mccarthytyler: Cool, sounds good.

PR #52, opened 2023-05-10
feat: effects api

mccarthytyler:
Start of the effects API.  Right now it requires `BaseEffect` from `@webex/web-media-effects` which brings in quite a few dependencies that we probably don't want.

### DIFF
--- a/src/media/local-stream.spec.ts
+++ b/src/media/local-stream.spec.ts
@@ -1,0 +1,117 @@
+import { BaseEffect } from '@webex/web-media-effects';
+import MediaStreamStub from '../mocks/media-stream-stub';
+import MediaStreamTrackStub from '../mocks/media-stream-track-stub';
+import { mocked } from '../mocks/mock';
+import { createMockedStream } from '../util/test-utils';
+import { LocalStream } from './local-stream';
+
+/**
+ * Create a mocked track effect.
+ *
+ * @returns A mocked TrackEvent.
+ */
+const createMockedTrackEffect = () => {
+  const effectTrack = mocked(new MediaStreamTrackStub());
+  const effect = {
+    dispose: jest.fn().mockResolvedValue(undefined),
+    load: jest.fn().mockResolvedValue(effectTrack),
+    on: jest.fn(),
+  };
+
+  return { effectTrack, effect };
+};
+
+/**
+ * A dummy LocalStream implementation so we can instantiate it for testing.
+ */
+class TestLocalStream extends LocalStream {
+  /**
+   * Fake method to get output stream.
+   *
+   * @returns The output stream.
+   */
+  getOutputStream() {
+    return this.outputStream;
+  }
+
+  /**
+   * Fake method to get output track.
+   *
+   * @returns The output track.
+   */
+  getOutputTrack() {
+    return this.outputStream.getTracks()[0];
+  }
+}
+
+describe('LocalStream', () => {
+  const mockStream = createMockedStream();
+  let localStream: LocalStream;
+  beforeEach(() => {
+    localStream = new TestLocalStream(mockStream);
+  });
+
+  it('should change the underlying track state based on being muted & unmuted', () => {
+    expect.assertions(2);
+    // Simulate the default state of the track's enabled state.
+    mockStream.getTracks()[0].enabled = true;
+
+    localStream.setMuted(true);
+    expect(mockStream.getTracks()[0].enabled).toBe(false);
+    localStream.setMuted(false);
+    expect(mockStream.getTracks()[0].enabled).toBe(true);
+  });
+});
+
+describe('LocalTrack addEffect', () => {
+  // eslint-disable-next-line jsdoc/require-jsdoc
+  const setup = () => {
+    const mockedStream = mocked(new MediaStreamStub());
+    const localStream = new TestLocalStream(mockedStream as unknown as MediaStream);
+    const { effectTrack, effect } = createMockedTrackEffect();
+    return { mockedStream, localStream, effectTrack, effect };
+  };
+
+  it('loads and uses the effect when there is no loading effect', async () => {
+    expect.hasAssertions();
+
+    const { localStream, effectTrack, effect } = setup();
+
+    const addEffectPromise = localStream.addEffect('test-effect', effect as unknown as BaseEffect);
+
+    await expect(addEffectPromise).resolves.toBeUndefined();
+    expect(localStream.getOutputTrack()).toBe(effectTrack);
+  });
+
+  it('does not use the effect when the loading effect is cleared during load', async () => {
+    expect.hasAssertions();
+
+    const { localStream, mockedStream, effect } = setup();
+
+    // Add effect and immediately dispose all effects to clear loading effects
+    const addEffectPromise = localStream.addEffect('test-effect', effect as unknown as BaseEffect);
+    await localStream.disposeEffects();
+
+    await expect(addEffectPromise).rejects.toThrow('not required after loading');
+    expect(localStream.getOutputStream()).toBe(mockedStream);
+  });
+
+  it('loads and uses the latest effect when the loading effect changes during load', async () => {
+    expect.hasAssertions();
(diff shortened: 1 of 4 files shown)


src/media/local-stream.ts
```diff
@@ -9,12 +10,30 @@ interface LocalStreamEvents {
   [LocalStreamEventNames.ConstraintsChange]: TypedEvent<() => void>;
 }
 
+export type TrackEffect = BaseEffect;
```
bbaldino, line 13: I think we can just use the name web-media-effects defines (BaseEffect).  If anything, maybe web-media-effects should rename BaseEffect to TrackEffect?

src/media/local-stream.ts
```diff
@@ -60,6 +79,50 @@ abstract class _LocalStream extends Stream {
...
+
+    // When the effect's track is updated, update the next effect or output stream.
+    effect.on(EffectEvent.TrackUpdated, (track: MediaStreamTrack) => {
+      const effectIndex = this.effects.findIndex((i) => i.name === name);
+      if (effectIndex === this.effects.length - 1) {
+        replaceTrack(this.outputStream, track);
+      } else {
+        this.effects[effectIndex + 1]?.effect.replaceInputTrack(track);
+      }
+    });
```
bbaldino, line 102: Do we need to do the 'pendingEffect' scheme that was recently added in webrtc-core main?

src/media/local-stream.ts
```diff
@@ -60,6 +79,50 @@ abstract class _LocalStream extends Stream {
...
   }
+
+  /**
+   * Adds an effect to a local stream.
+   *
+   * @param name - The name of the effect.
+   * @param effect - The effect to add.
+   */
+  async addEffect(name: string, effect: TrackEffect): Promise<void> {
+    const outputTrack = await effect.load(this.outputStream.getTracks()[0]);
```
bbaldino, line 90: Should we first check if an effect with this name is already in `this.effects`?

src/media/local-stream.ts
```diff
@@ -60,6 +79,50 @@ abstract class _LocalStream extends Stream {
...
+    this.effects.push({ name, effect });
+    replaceTrack(this.outputStream, outputTrack);
+
+    // When the effect's track is updated, update the next effect or output stream.
+    effect.on(EffectEvent.TrackUpdated, (track: MediaStreamTrack) => {
+      const effectIndex = this.effects.findIndex((i) => i.name === name);
```
bbaldino, line 96: nit: use `e` instead of `i` here? `i` makes me think it's an index.

src/media/local-stream.ts
```diff
@@ -60,6 +81,65 @@ abstract class _LocalStream extends Stream {
...
+    // Check that the loaded effect is the latest one and dispose if not
+    if (this.loadingEffects.indexOf(name) > -1) {
+      await effect.dispose();
+      throw new Error(`Effect "${name}" not required after loading`);
+    }
+
+    // Prevent loading multiple effects wth the same name
+    if (this.effects.findIndex((e) => e.name === name)) {
+      await effect.dispose();
+      throw new Error(`Effect already exists with name "${name}"`);
```
bbaldino, line 101: According to the web client team, there's a use case to replace an earlier version of an effect with a newer one (which may have different params)--I think that's what the current webrtc-core supports.  If we want to enforce they dispose the old one first, we should check in with them.

src/media/local-stream.ts
```diff
@@ -9,12 +10,32 @@ interface LocalStreamEvents {
...
+
 /**
  * A stream which originates on the local device.
  */
 abstract class _LocalStream extends Stream {
   [LocalStreamEventNames.ConstraintsChange] = new TypedEvent<() => void>();
 
+  private effects: EffectItem[] = [];
+
+  private loadingEffects: string[] = [];
```
bbaldino, line 37: Because order isn't important here, I think this could be a map keyed by the effect name, which would make the lookups a bit easier.

src/media/local-stream.ts
mccarthytyler: Ah I must have misread.  For now I can replicate their implementation.

src/media/local-stream.ts
mccarthytyler: Added.

src/media/local-stream.ts
mccarthytyler: Updated.

src/media/local-stream.ts
mccarthytyler: Updated.

src/media/local-stream.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+    localStream = new TestLocalStream(mockStream);
+  });
+
+  it('should change the underlying track state based on being muted & unmuted', () => {
+    expect.assertions(2);
+    // Simulate the default state of the track's enabled state.
+    mockStream.getTracks()[0].enabled = true;
+
+    localStream.setMuted(true);
+    expect(privateAccess(localStream).outputStream.getTracks()[0].enabled).toBe(false);
```
bbaldino, line 51: can't we verify these on `mockStream` instead of using `privateAccess`?

src/media/local-stream.ts
```diff
@@ -130,7 +130,7 @@ abstract class _LocalStream extends Stream {
    */
   async disposeEffects(): Promise<void> {
     // Dispose of any effects currently in use
-    if (this.effects.length > 0) {
+    if (this.loadingEffects.size > 0) {
```
bbaldino, line 133: I think we do want this to be `this.effects`?

src/media/local-stream.spec.ts
mccarthytyler: Updated.

src/media/local-stream.ts
mccarthytyler: Updated.

src/media/local-stream.spec.ts
```diff
@@ -0,0 +1,108 @@
...
+
+  it('loads and uses the effect when there is no loading effect', async () => {
+    expect.hasAssertions();
+
+    const { localStream, effectTrack, effect } = setup();
+
+    const addEffectPromise = localStream.addEffect('test-effect', effect as unknown as BaseEffect);
+
+    await expect(addEffectPromise).resolves.toBeUndefined();
+    expect(privateAccess(localStream).outputStream.getTracks()[0]).toBe(effectTrack);
```
bbaldino, line 74: I think these uses of privateAccess can be changed too?

bbaldino (commented):
one more nit on the tests, but otherwise lgtm

bbaldino (commented):
one more nit on the tests, but otherwise lgtm

src/media/local-stream.spec.ts
bbaldino: i think this change may have gotten lost?

mccarthytyler: I changed the above code to:

```ts
mockStream.getTracks()[0].enabled = true;

localStream.setMuted(true);
expect(mockStream.getTracks()[0].enabled).toBe(false);
localStream.setMuted(false);
expect(mockStream.getTracks()[0].enabled).toBe(true);
```

src/media/local-stream.spec.ts
mccarthytyler: In these cases I think we want to make sure the output stream has the track that the effect is outputting

bbaldino: ok...how about just having `TestLocalStream` expose it?

bbaldino: Though, to be honest, I can't remember why `Stream` doesn't expose a getter for `outputStream`, I'm pretty sure we'll end up with one (even it's just an abstract getter at the `Stream` level)

mccarthytyler: I ended up having `TestLocalStream` expose it, I can go back and update if/when the getter is added.  Is that something we want to add on this PR?

bbaldino: Ok cool. Nah we'll do that later

PR #56, opened 2023-06-30
feat: limit appliable constraints

brycetham:
This PR limits the constraints that can be passed to the `applyConstraints` method in both `LocalAudioStream` and `LocalVideoStream`.


src/media/local-audio-stream.ts
```diff
@@ -1,20 +1,10 @@
 import { logger } from '../util/logger';
 import { LocalStream, LocalStreamEventNames } from './local-stream';
 
-// TODO: for applyConstraints, we only support things which can be changed on the
-// track itself.  This would not include 'deviceId', and maybe more of these values.
-// Might need to define one AudioConstraints for use with getUserMedia, and refine
-// it even further for applyConstraints here.
-export type AudioConstraints = Pick<
```
bbaldino, line 8: Do we have these defined somewhere? The 'full set' should be usable in the getUserMedia calls

brycetham: We do have `AudioDeviceConstraints` and `VideoDeviceConstraints` defined in `device-management.ts`, if that's what you're asking.

brycetham: Unrelated, but just noticed that we're not picking from `MediaTrackConstraints` when defining `VideoDeviceConstraints` like we do with `AudioDeviceConstraints`. I can make that change real fast.

bbaldino: ok cool

bbaldino (approved):

PR #57, opened 2023-07-13
feat: listen to track events and additional updates to stream classes

brycetham:
This PR makes several additional changes to the stream classes, including listening to media stream track events in `Stream`, adding more abstract methods, and updating how `inputTrack` and `outputTrack` are used in `LocalStream`.

### DIFF
--- a/src/media/local-stream.ts
+++ b/src/media/local-stream.ts
@@ -90,23 +90,30 @@ abstract class _LocalStream extends Stream {
   }
 
   /**
-   * Change the track of the output stream to a different track. If the new track is the input
-   * track, then set the output stream to be the input stream. If the input and output tracks are
-   * currently the same, then set the output stream to a new stream with the new track. Otherwise,
-   * remove the old track from the output stream and add the new one.
+   * Change the track of the output stream to a different track.
+   *
+   * Note: this method assumes and enforces that if both input and output streams have the same
+   * track, then they must also be the same stream.
    *
    * @param newTrack - The track to be used in the output stream.
    */
   private changeOutputTrack(newTrack: MediaStreamTrack): void {
-    if (this.inputTrack.id === newTrack.id) {
-      this._outputStream = this.inputStream;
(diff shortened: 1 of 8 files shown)


src/media/local-stream.ts
```diff
@@ -78,44 +68,53 @@ abstract class _LocalStream extends Stream {
...
+   * @param newTrack - The track to be used in the output stream.
    */
-  private replaceTrack(newTrack: MediaStreamTrack): void {
-    this.outputStream.removeTrack(this.outputTrack);
-    this.outputStream.addTrack(newTrack);
+  private changeOutputTrack(newTrack: MediaStreamTrack): void {
+    if (this.inputTrack.id === newTrack.id) {
+      this._outputStream = this.inputStream;
+    } else if (this.inputTrack.id === this.outputTrack.id) {
+      this._outputStream = new MediaStream([newTrack]);
```
bbaldino, line 104: We have to preserve the output stream, since that's what callers will have a handle to.  So we actually need to create a new stream that'll be used as the input stream, and move the current input/output track to that, and then add the new track to the existing outputstream.  (Tweak the comment as well)

bbaldino: I'd think enabling background blur or something should show that this doesn't work: the local rendering of the video wouldn't show the blur since the outputstream changes?

src/media/local-stream.ts
```diff
@@ -90,23 +90,30 @@ abstract class _LocalStream extends Stream {
...
+
       this.outputStream.removeTrack(this.outputTrack);
       this.outputStream.addTrack(newTrack);
+
+      // input track is same as new output track, so set the streams to be the same
+      if (this.inputTrack.id === this.outputTrack.id) {
+        this.inputStream = this.outputStream;
+      }
+
+      this[LocalStreamEventNames.OutputTrackChange].emit(newTrack);
```
bbaldino: I think be really verbose/descriptive with the comments here, the logic is pretty dense.

bbaldino (approved):

PR #58, opened 2023-07-18
feat: convert track-based classes into stream-based classes

brycetham:
BREAKING CHANGE: This PR converts the track-based classes into stream-based classes. The new class hierarchy now looks like this:

- `Stream`
    - `LocalStream` (previously `LocalTrack`)
        - `LocalAudioStream` 
            - `LocalMicrophoneStream` (previously `LocalMicrophoneTrack`)
            - `LocalSystemAudioStream` (previously `LocalSystemAudioTrack`)
        - `LocalVideoStream` 
            - `LocalCameraStream` (previously `LocalCameraTrack`)
            - `LocalDisplayStream` (previously `LocalDisplayTrack`)
    - `RemoteStream`

The new classes emit `TypedEvent`s from ts-events. The device management API has been updated to return these new classes.

More details can be found here: https://confluence-eng-gpk2.cisco.com/conf/x/yyoCFw


src/media/local-stream.ts
```diff
@@ -0,0 +1,194 @@
...
+
+      this[LocalStreamEventNames.OutputTrackChange].emit(newTrack);
+    }
+  }
+
+  /**
+   * @inheritdoc
+   */
+  stop(): void {
+    this.inputTrack.stop();
```
marcin-bazyl, line 126: do we need to stop also the outputTrack?

brycetham: Yeah we should. Good catch, will fix it.

bbaldino (approved):
:tada:

PR #61, opened 2023-07-28
feat: add generics to device management

brycetham:
This PR adds generics to device management `create*Stream` functions in order to support using custom stream classes in these methods. It also adds a new API, `createCameraAndMicrophoneStreams`, to create both a camera stream and a microphone stream with a single `getUserMedia` call.

As part of this change, I've also slightly modified the `ErrorTypes` to have a generic `CREATE_STREAM_FAILED` error so it can be used in all cases in which `getUserMedia` or `getDisplayMedia` fail.

This completes [SPARK-448441](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-448441) and [SPARK-448443](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-448443).

### DIFF
--- a/src/device/device-management.ts
+++ b/src/device/device-management.ts
@@ -7,9 +7,11 @@ import { VideoContentHint } from '../media/local-video-stream';
 
 export enum ErrorTypes {
   DEVICE_PERMISSION_DENIED = 'DEVICE_PERMISSION_DENIED',
-  CREATE_CAMERA_STREAM_FAILED = 'CREATE_CAMERA_STREAM_FAILED',
-  CREATE_MICROPHONE_STREAM_FAILED = 'CREATE_MICROPHONE_STREAM_FAILED',
+  CREATE_STREAM_FAILED = 'CREATE_CAMERA_STREAM',
 }
+
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type Constructor<T> = new (...args: any[]) => T;
 
 /**
  * Represents a WCME error, which contains error type and error message.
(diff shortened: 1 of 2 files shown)


src/device/device-management.ts
```diff
@@ -7,10 +7,12 @@ import { VideoContentHint } from '../media/local-video-stream';
...
-  CREATE_CAMERA_STREAM_FAILED = 'CREATE_CAMERA_STREAM_FAILED',
-  CREATE_MICROPHONE_STREAM_FAILED = 'CREATE_MICROPHONE_STREAM_FAILED',
+  CREATE_STREAM_FAILED = 'CREATE_CAMERA_STREAM',
 }
 
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
+type Constructor<T> = new (...args: any[]) => T;
+
 /**
  * Represents a WCME error, which contains error type and error message.
```
brycetham, line 17: We can do this in a later PR, but I think it's kind of weird that we're referencing WCME here, and naming the error class `WcmeError`, especially since WCME doesn't actually call any of the device management APIs.

src/device/device-management.ts
```diff
@@ -54,56 +56,109 @@ export type VideoDeviceConstraints = Pick<
...
+ * @param constraints.audio - Audio device constraints.
+ * @returns A Promise that resolves to a LocalCameraStream and a LocalMicrophoneStream or an error.
+ */
+export async function createCameraAndMicrophoneStreams<
+  T extends LocalCameraStream,
+  U extends LocalMicrophoneStream
+>(
+  cameraStreamConstructor: Constructor<T>,
+  microphoneStreamConstructor: Constructor<U>,
+  constraints?: { video?: VideoDeviceConstraints; audio?: AudioDeviceConstraints }
```
arun3528, line 118: nice

arun3528 (approved):

PR #62, opened 2023-09-29
feat: add id, readyState, and getAllEffects to stream classes

brycetham:
This PR adds an `id` getter to the `Stream` class, as well as a `readyState` getter and a `getAllEffects` method to the `LocalStream` class.

### DIFF
--- a/src/media/stream.ts
+++ b/src/media/stream.ts
@@ -1,4 +1,5 @@
 import { AddEvents, TypedEvent, WithEventsDummyType } from '@webex/ts-events';
+import { v4 as uuid } from 'uuid';
 
 export enum StreamEventNames {
   MuteStateChange = 'mute-state-change',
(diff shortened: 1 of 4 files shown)


src/media/stream.ts
```diff
@@ -14,6 +15,8 @@ interface StreamEvents {
  * Base stream class.
  */
 abstract class _Stream {
+  readonly id = uuid();
```
bbaldino, line 18: Is there a reason not to use [`MediaStream::id`](https://developer.mozilla.org/en-US/docs/Web/API/MediaStream/id)?

brycetham: I thought originally that maybe outputStream would change, but I see now that we had explicitly made it so that the stream itself would always remain the same, so yeah I think I can change the id getter to return `this.outputStream.id`.

bbaldino (approved):

PR #63, opened 2023-10-12
fix: throw error in setLocalDescription if media line is missing codecs

brycetham:
This PR fixes an issue in which Firefox will allow `setLocalDescription` to be called even when all the codecs have been removed from the media line as part of [SPARK-465378](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-465378). This violates RFC8866, which states that at least one codec is required.

A bug has been filed to Firefox about this: https://bugzilla.mozilla.org/show_bug.cgi?id=1857612

Until Firefox fixes this bug, the fix from our end is to add a check in `setLocalDescription` to check the number of fields in the media line. This is similar to what Google does: https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/pc/webrtc_sdp.cc;l=2682;drc=f5bdc89c7395ed24f1b8d196a3bdd6232d5bf771

If the number of fields is less than 4 (i.e. the codecs are missing), then we throw an error.

### DIFF
--- a/package.json
+++ b/package.json
@@ -107,6 +107,7 @@
   },
   "dependencies": {
     "@webex/ts-events": "^1.1.0",
+    "@webex/web-capabilities": "^1.0.0",
     "@webex/web-media-effects": "^2.7.0",
     "events": "^3.3.0",
     "js-logger": "^1.6.1",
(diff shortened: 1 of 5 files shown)


brycetham:
Keep this as draft until https://github.com/webex/web-capabilities/pull/3 is merged.

bbaldino (approved):

PR #67, opened 2023-12-12
fix: handle invalid m-line with trailing whitespace

brycetham:
This PR fixes the case in which an invalid m-line with a trailing whitespace may be considered to be valid. For example, "m=video 9 UDP/TLS/RTP/SAVPF " should be throwing an error, but the trailing space causes webrtc-core to think it is valid.


bbaldino (approved):
nice find

PR #68, opened 2024-01-08
feat: update effects APIs

brycetham:
This PR makes API changes to the `LocalStream` class regarding effects. The main change is eliminating the need to pass a `name` param to the `addEffect` method, since the effect object has its own identifiers as of https://github.com/webex/web-media-effects/pull/59.

The following APIs have been added:
- `getEffectById`
- `getEffectsByKind`

The following API has been removed:
- `getEffect`

The following API has been renamed:
- `getAllEffects` -> `getEffects` (I think this new name makes sense with the new APIs; it matches the built-in `MediaStream` APIs, which have `getTracks` and `getTrackById`, for example)


src/media/local-stream.ts
```diff
@@ -141,55 +144,80 @@ abstract class _LocalStream extends Stream {
...
-   * @param name - The name of the effect.
    * @param effect - The effect to add.
    */
-  async addEffect(name: string, effect: TrackEffect): Promise<void> {
-    // Load the effect
-    this.loadingEffects.set(name, effect);
-    const outputTrack = await effect.load(this.outputTrack);
+  async addEffect(effect: TrackEffect): Promise<void> {
+    // Check if the effect has already been added.
+    if (this.effects.includes(effect)) {
```
bbaldino, line 151: What does this end up using to compare equality?

src/media/local-stream.ts
```diff
@@ -141,55 +144,80 @@ abstract class _LocalStream extends Stream {
...
-    // Check that the loaded effect is the latest one and dispose if not
-    if (effect !== this.loadingEffects.get(name)) {
+    // Load the effect. Because loading is asynchronous, keep track of the loading effects.
+    this.loadingEffects.set(effect.kind, effect);
+    await effect.load(this.outputTrack);
+
+    // After loading, check whether or not we still want to use this effect. If another effect of
+    // the same kind was added while this effect was loading, we only want to use the latest effect,
+    // so dispose this one. If the effects list was cleared while this effect was loading, also
+    // dispose it.
```
bbaldino, line 165: It feels a bit strange that we'll allow overriding an affect of the same type that was added previously _if_ it didn't finish loading yet, but we error on adding an effect of the same type after it's been loaded.  I think I'd expect that subsequent additions of the same type of effect (without removing the previous one) should either always work or always fail?

src/media/local-stream.ts
```diff
@@ -141,55 +144,80 @@ abstract class _LocalStream extends Stream {
...
+    return this.effects.find((effect) => effect.id === id);
+  }
+
+  /**
+   * Get all the effects from the effects list with the given kind.
+   *
+   * @param kind - The kind of the effects you want to get.
+   * @returns A list of effects.
+   */
+  getEffectsByKind(kind: string): TrackEffect[] {
```
bbaldino, line 211: This returning an array seems inconsistent with the fact that we only allow one instance of each effect type

src/media/local-stream.ts
brycetham: It just checks to see if the instances match, but I do think explicitly comparing `id` or `kind` would be better, so I can make that change.

src/media/local-stream.ts
brycetham: The previous behavior was to override the effect if it didn't finish loading, so I think we should keep that behavior (since the client does depend on it). We didn't handle adding a duplicate effect after it has loaded previously, so I can change it so that it will also override instead of error.

src/media/local-stream.ts
brycetham: Sure, will change it.

arun3528: just curious , if we knew there is only 3 kind of effects why didn't we have some function like addVirtualBackgroundEffect rather have names passed in to the effects ?

bbaldino: Well there are 3 effects _now_, but we want to support any kind of effect in the future as opposed to hard-coding the current ones in.

src/media/local-stream.ts
```diff
@@ -148,11 +149,8 @@ abstract class _LocalStream extends Stream {
    */
   async addEffect(effect: TrackEffect): Promise<void> {
     // Check if the effect has already been added.
-    if (this.effects.includes(effect)) {
-      throw new WebrtcCoreError(
-        WebrtcCoreErrorType.ADD_EFFECT_FAILED,
-        `Effect ${effect.id} has already been added to this stream.`
-      );
+    if (this.effects.some((e) => e.id === effect.id)) {
```
bbaldino, line 152: So this now checks if the _same_ effect instance is being added, but I think we need to check for an existing effect of the same kind to replace it with this one?

bbaldino: Oh, I didn't notice that all the code below is in this same method, I see there's handling there  for this.

brycetham: There's logic starting at line 205 that handles this case.

bbaldino (approved):

PR #69, opened 2024-01-18
fix: check muted property for mute getter

brycetham:
This PR also checks the mute property of the track when calling `LocalStream.muted`. This more accurately reflects the mute state of the track; for example, when sharing a window, the track gets muted by the browser when the shared window is minimized.


bbaldino (dismissed):

brycetham:
Holding off on merging this until I make some changes regarding the `MuteStateChange` event. We should only fire this if the mute state actually changes, taking into account both `muted` and `enabled` properties.

bbaldino (approved):

PR #70, opened 2024-01-23
feat: add custom method toJSON for _LocalStream

antsukanova:
The main purpose of this update is to make better work with LocalStream and smoother debugging for [webex-js-sdk](https://github.com/webex/webex-js-sdk). Mainly it should get rid of custom overrides for objects and give more useful information in better structure. See current overrides here https://github.com/webex/webex-js-sdk/pull/3148/commits/4089648179ee2566cab1e7e718c6ac5908c83eb7

This PR consists of the following updates: 
- update for `_LocalStream` object and add custom `toJSON` method
- added new test cases for `_LocalStream`

### DIFF
--- a/src/media/local-stream.ts
+++ b/src/media/local-stream.ts
@@ -291,13 +291,13 @@ abstract class _LocalStream extends Stream {
    */
   toJSON() {
     return {
+      muted: this.muted,
+      enabled: this.inputTrack.enabled,
+      label: this.label,
+      readyState: this.readyState,
       inputStream: {
         active: this.inputStream.active,
         id: this.inputStream.id,
-        muted: this.muted,
-        label: this.label,
-        enabled: this.inputTrack.enabled,
-        readyState: this.readyState,
       },
       outputStream: {
         active: this.outputStream.active,
--- a/src/media/local-stream.ts
+++ b/src/media/local-stream.ts
@@ -284,6 +284,32 @@ abstract class _LocalStream extends Stream {
   }
 
   /**
+   * Method to serialize data about input, output streams
+   * and also effects from _LocalStream.
+   *
+   * @returns - A JSON-compatible object representation with data from _LocalStream.
+   */
+  toJSON() {
+    return {
+      inputStream: {
+        active: this.inputStream.active,
+        id: this.inputStream.id,
+      },
+      outputStream: {
+        active: this.outputStream.active,
+        id: this.outputStream.id,
+      },
+      effects: this.effects.map((effect) => {
+        return {
+          id: effect.id,
+          kind: effect.kind,
+          isEnabled: effect.isEnabled,
+        };
+      }),
+    };
+  }
+
+  /**
    * Cleanup the local effects.
    */
   async disposeEffects(): Promise<void> {
--- a/src/media/local-stream.ts
+++ b/src/media/local-stream.ts
@@ -292,12 +292,13 @@ abstract class _LocalStream extends Stream {
   toJSON() {
     return {
       muted: this.muted,
-      enabled: this.inputTrack.enabled,
       label: this.label,
       readyState: this.readyState,
       inputStream: {
         active: this.inputStream.active,
         id: this.inputStream.id,
+        enabled: this.inputTrack.enabled,
+        muted: this.inputTrack.muted,
       },
       outputStream: {
         active: this.outputStream.active,
--- a/src/media/local-stream.ts
+++ b/src/media/local-stream.ts
@@ -294,6 +294,10 @@ abstract class _LocalStream extends Stream {
       inputStream: {
         active: this.inputStream.active,
         id: this.inputStream.id,
+        muted: this.muted,
+        label: this.label,
+        enabled: this.inputTrack.enabled,
+        readyState: this.readyState,
       },
       outputStream: {
         active: this.outputStream.active,
--- a/src/media/local-stream.ts
+++ b/src/media/local-stream.ts
@@ -285,9 +285,9 @@ abstract class _LocalStream extends Stream {
 
   /**
    * Method to serialize data about input, output streams
-   * and also effects from _LocalStream.
+   * and also effects from LocalStream.
    *
-   * @returns - A JSON-compatible object representation with data from _LocalStream.
+   * @returns - A JSON-compatible object representation with data from LocalStream.
    */
   toJSON() {
     return {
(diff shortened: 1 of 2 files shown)


brycetham (commented):
Code looks fine, but I'll defer to @marcin-bazyl to see if the output of the `toJSON` function matches his expectations. I noticed the current workaround in the SDK includes `label`, `readyState`, and `enabled` (though I'd argue we should use `muted` instead for the last one).

src/media/local-stream.ts
```diff
@@ -283,6 +283,32 @@ abstract class _LocalStream extends Stream {
     return this.effects;
   }
 
+  /**
+   * Method to serialize data about input, output streams
+   * and also effects from _LocalStream.
+   *
+   * @returns - A JSON-compatible object representation with data from _LocalStream.
+   */
+  toJSON() {
```
bbaldino, line 292: nit: let's do `toJson`

src/media/local-stream.ts
```diff
@@ -283,6 +283,32 @@ abstract class _LocalStream extends Stream {
...
+    return {
```
bbaldino, line 293: Maybe another nitpick, but this is returning an objec, not "json", right?  I think if this method is going to be called `toJson` we should `JSON.stringify` and return a `String`.

antsukanova: @bbaldino yes and with your idea, it requires you to manually call your custom method each time you need it, and it won't be used automatically by `JSON.stringify()`. That's more transparent. But I also want to ask @marcin-bazyl will it be convenient to use it in this way?

src/media/local-stream.ts
marcin-bazyl: `toJson` won't work, it has to be `toJSON`, because that's what `JSON.stringify()` calls, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/stringify

src/media/local-stream.ts
marcin-bazyl: @bbaldino  the whole point of this jira is to make JSON.stringify() work, not to invent new methods. Any library/component that uses webrtc-core should be free to call JSON.stringify() on instances of webrtc-core classes like LocalStream and that should not cause the browser to hang up (that's what's currently happening).

marcin-bazyl:
> Code looks fine, but I'll defer to @marcin-bazyl to see if the output of the `toJSON` function matches his expectations. I noticed the current workaround in the SDK includes `label`, `readyState`, and `enabled` (though I'd argue we should use `muted` instead for the last one).

yeah, good point, it would be good to include `label`, `readyState`, `enabled` and `muted` from the inputTrack

src/media/local-stream.ts
bbaldino: Oh, sorry didn't realize this was an automatic hook for JSON.stringify. 👍

src/media/local-stream.ts
bbaldino: 👍

src/media/local-stream.ts
```diff
@@ -283,6 +283,36 @@ abstract class _LocalStream extends Stream {
...
+   * @returns - A JSON-compatible object representation with data from LocalStream.
+   */
+  toJSON() {
+    return {
+      inputStream: {
+        active: this.inputStream.active,
+        id: this.inputStream.id,
+        muted: this.muted,
+        label: this.label,
+        enabled: this.inputTrack.enabled,
```
antsukanova, line 299: @bbaldino @brycetham not sure that for one case here it is useful to have a getter for `enabled`. So right now put it via `inputTrack`. Let me know if you have any other thoughts

antsukanova:
@marcin-bazyl @bbaldino @brycetham here is an example from the console of how it looks like for the test-app. Please let me know @marcin-bazyl  if it is what you need.
Screenshot 2024-01-25 at 13 40 18

src/media/local-stream.ts
brycetham: Yeah I wouldn't add a getter for `enabled` unless there is a need for it outside this function. Actually, I would argue that we shouldn't include `enabled` here at all, since `muted` already takes `enabled` into account.

I would also probably move `muted`, `label`, and `readyState` out of `inputStream` and put it at the same level as `inputStream`, `outputStream`, and `effects`. These things, in my opinion, apply to the whole `LocalStream`, not just the input (hence why we have explicitly defined getters for all of them).

marcin-bazyl: for debugging purposes it's better to keep both `muted` and `enabled`

brycetham: Ok, but let's keep it separate from the overall `LocalStream` mute state. So I would suggest something like this:

```
{
  muted: this.muted,
  inputStream: {
    muted: this.inputTrack.muted,
    enabled: this.inputTrack.enabled,
    ...
  },
  ...
}
```

antsukanova: @brycetham updated recording your advice, thanks
@bbaldino @marcin-bazyl FYI

src/media/local-stream.ts
```diff
@@ -283,6 +283,36 @@ abstract class _LocalStream extends Stream {
...
+   */
+  toJSON() {
+    return {
+      muted: this.muted,
+      enabled: this.inputTrack.enabled,
+      label: this.label,
+      readyState: this.readyState,
+      inputStream: {
+        active: this.inputStream.active,
+        id: this.inputStream.id,
```
brycetham, line 300: ...and add `enabled: this.inputTrack.enabled` and `muted: this.inputTrack.muted` here. This way, we can tell the `enabled` and `muted` state of the input track apart from the overall `muted` state of the `LocalStream`.

src/media/local-stream.ts
```diff
@@ -283,6 +283,36 @@ abstract class _LocalStream extends Stream {
...
+  /**
+   * Method to serialize data about input, output streams
+   * and also effects from LocalStream.
+   *
+   * @returns - A JSON-compatible object representation with data from LocalStream.
+   */
+  toJSON() {
+    return {
+      muted: this.muted,
+      enabled: this.inputTrack.enabled,
```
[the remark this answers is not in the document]
antsukanova: @brycetham I understand your idea now. It sounds logical. Thank you for clarifying it for me. Will update

src/media/local-stream.ts
antsukanova: @brycetham updated this part, thanks)

brycetham (approved):
Looks good to me 👍

marcin-bazyl (approved):

antsukanova:
Thanks, guys, for your review. Merge it 👍

PR #71, opened 2024-02-08
feat: set/create events

SomeBody16:
Added events that will be called on success operations on offer/answer/description

This includes
- `CreateOfferOnSuccess` - when `createOffer` succeeded
- `CreateAnswerOnSuccess` - when `createAnswer` succeeded
- `SetLocalDescriptionOnSuccess` - when `setLocalDescription` succeeded
- `SetRemoteDescriptionOnSuccess` - when `setRemoteDescription` succeeded

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@webex/webrtc-core",
-  "version": "2.4.0",
+  "version": "2.4.1",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
   "browser": "dist/umd/index.js",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@webex/webrtc-core",
-  "version": "2.4.1",
+  "version": "2.4.0",
   "main": "dist/cjs/index.js",
   "module": "dist/esm/index.js",
   "browser": "dist/umd/index.js",
--- a/src/peer-connection.spec.ts
+++ b/src/peer-connection.spec.ts
@@ -273,7 +273,7 @@ describe('PeerConnection', () => {
       mockCreateRTCPeerConnection.mockReturnValueOnce(mockPc as unknown as RTCPeerConnection);
       pc = new PeerConnection();
       createOfferSpy = jest.spyOn(pc, 'createOffer');
-      pc.on(PeerConnection.Events.CreateOffer, callback);
+      pc.on(PeerConnection.Events.CreateOfferOnSuccess, callback);
     });
 
     it('should emit event when createOffer called', async () => {
@@ -317,7 +317,7 @@ describe('PeerConnection', () => {
       });
       setLocalDescriptionSpy = jest.spyOn(mockPc, 'setLocalDescription');
       pc = new PeerConnection();
-      pc.on(PeerConnection.Events.SetLocalDescription, callback);
+      pc.on(PeerConnection.Events.SetLocalDescriptionOnSuccess, callback);
     });
 
     it('sets the local description with an SDP offer', async () => {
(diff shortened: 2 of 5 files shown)


src/peer-connection.ts
```diff
@@ -24,14 +24,22 @@ type IceGatheringStateChangeEvent = {
   target: EventTarget | null;
 };
 
+type SdpChangeEvent = {
+  type: 'offer' | 'answer';
+  munged: boolean;
+  sdp?: string;
+};
```
brycetham, line 31: A couple of things here.

1. I think I prefer having separate events for each `createOffer`, `setLocalDescription`, and `setRemoteDescription`. It may not always be the case that the user will want to listen to all of these events, so it would be better for them to pick and choose which ones to listen to.
2. Remember that "munge" refers to modifying the SDP. But actually, we don't know if the SDP when calling `setLocalDescription` or `setRemoteDescription` will be modified. For example, someone might call `createOffer` and then `setLocalDescription` right away without munging the SDP.

src/peer-connection.ts
```diff
@@ -215,6 +229,12 @@ class PeerConnection extends EventEmitter<PeerConnectionEventHandlers> {
         });
     }
 
+    this.emit(PeerConnection.Events.SdpChange, {
+      type: 'offer',
+      munged: true,
+      sdp: description?.sdp,
+    });
+
     return this.pc.setLocalDescription(description);
```
brycetham, line 238: It's possible for the call to `this.pc.setLocalDescription` to fail, in which case we wouldn't want to emit an event saying the SDP changed. We probably only want to emit this event after `this.pc.setLocalDescription` succeeds.

src/peer-connection.ts
```diff
@@ -230,6 +250,12 @@ class PeerConnection extends EventEmitter<PeerConnectionEventHandlers> {
   async setRemoteDescription(
     description: RTCSessionDescription | RTCSessionDescriptionInit
   ): Promise<void> {
+    this.emit(PeerConnection.Events.SdpChange, {
+      type: 'answer',
+      munged: true,
+      sdp: description.sdp,
+    });
+
     return this.pc.setRemoteDescription(description);
```
brycetham, line 259: Same here.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@webex/webrtc-core",
-  "version": "2.4.0",
+  "version": "2.4.1",
```
brycetham, line 3: No need to change this, it will be changed automatically when we generate a release :)

bbaldino:
I think we should consider here what the native events look like and think about being consistent:  the real events don't fire an event with the SDP on `createOffer`, for example, but instead `createOfferOnSuccess`, I believe.  Same for `setLocalDescription`, etc.

src/peer-connection.spec.ts
```diff
@@ -247,17 +248,76 @@ describe('PeerConnection', () => {
...
+    let mockPc: MockedObjectDeep<RTCPeerConnectionStub>;
     let setLocalDescriptionSpy: jest.SpyInstance;
+    const callback = jest.fn();
     let pc: PeerConnection;
 
     beforeEach(() => {
       jest.clearAllMocks();
       mockPc = mocked(new RTCPeerConnectionStub(), true);
       mockCreateRTCPeerConnection.mockReturnValueOnce(mockPc as unknown as RTCPeerConnection);
+      mockPc.setLocalDescription.mockImplementation((description) => {
```
brycetham, line 309: `description` may not be the best param name for this, since you typically don't pass an SDP into a call to `createOffer`. I also think it may be confusing to throw if no param is provided, since technically `createOffer` can be called with no params (same with `setLocalDescription`).

Wondering if it might be cleaner to mock `createOffer` so it always succeeds here, and then overwrite the mock to fail only in the test cases where we do want it to fail?

brycetham (commented):
Looks good! Just one small comment on the unit tests.

Also, for completeness, maybe we should add an event for `createAnswerOnSuccess` too. We don't use that API much but we might as well add it while we're working on these other events.

Let's also update the PR description to reflect your recent changes, in case we want to refer back to this PR in the future.

brycetham (approved):

PR #72, opened 2024-02-28
feat: separate API for enabled and muted states

brycetham:
This PR updates the mute events and getters to make it clear whether the user muted the stream or the system muted the stream as part of [SPARK-493485](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-493485). The main changes are:

`Stream`:
- Remove event `MuteStateChange`. This will be replaced with new events in `LocalStream`/`RemoteStream`.

`LocalStream`:
- Add event `UserMuteStateChange`, emits a boolean with the new user mute state.
- Add event `SystemMuteStateChange`, emits a boolean with the new system mute state.
- Getters `userMuted`/`systemMuted`
- `muted` will be in `LocalStream` only, gets the overall mute state (taking into consideration both `userMuted` and `systemMuted`)

`RemoteStream`:
- Add event `MediaStateChange`, emits either "started" or "stopped".
- Remove getter `muted` and replace with getter `mediaState`, returns "started" or "stopped".

### DIFF
--- a/src/media/local-stream.ts
+++ b/src/media/local-stream.ts
@@ -2,7 +2,7 @@ import { AddEvents, TypedEvent, WithEventsDummyType } from '@webex/ts-events';
 import { BaseEffect, EffectEvent } from '@webex/web-media-effects';
 import { WebrtcCoreError, WebrtcCoreErrorType } from '../errors';
 import { logger } from '../util/logger';
-import { Stream, StreamEventNames } from './stream';
+import { MuteStateChangeReason, Stream, StreamEventNames } from './stream';
 
 export type TrackEffect = BaseEffect;
 
@@ -88,6 +88,9 @@ abstract class _LocalStream extends Stream {
   /**
    * Set the mute state of this stream.
    *
+   * Note: This sets the user-toggled mute state, equivalent to changing the "enabled" state of the
+   * track. It is separate from the browser-toggled mute state.
+   *
    * @param isMuted - True to mute, false to unmute.
    */
   setMuted(isMuted: boolean): void {
@@ -95,7 +98,7 @@ abstract class _LocalStream extends Stream {
       this.inputTrack.enabled = !isMuted;
       // setting `enabled` will not automatically emit MuteStateChange, so we emit it here
       if (!this.inputTrack.muted) {
-        this[StreamEventNames.MuteStateChange].emit(isMuted);
+        this[StreamEventNames.MuteStateChange].emit(isMuted, MuteStateChangeReason.ByUser);
       }
     }
   }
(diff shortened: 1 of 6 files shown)


src/media/stream.ts
```diff
@@ -5,8 +5,15 @@ export enum StreamEventNames {
   Ended = 'stream-ended',
 }
 
+export enum MuteStateChangeReason {
```
bbaldino, line 8: A bit of a nitpick, but I wonder if we should re-phrase this as "MuteStateChangeSource" and the values be more "plain nouns" (`User`, `Browser`).

src/media/local-stream.ts
```diff
@@ -88,14 +88,17 @@ abstract class _LocalStream extends Stream {
...
+   *
    * @param isMuted - True to mute, false to unmute.
    */
   setMuted(isMuted: boolean): void {
     if (this.inputTrack.enabled === isMuted) {
       this.inputTrack.enabled = !isMuted;
       // setting `enabled` will not automatically emit MuteStateChange, so we emit it here
       if (!this.inputTrack.muted) {
-        this[StreamEventNames.MuteStateChange].emit(isMuted);
+        this[StreamEventNames.MuteStateChange].emit(isMuted, MuteStateChangeReason.ByUser);
```
marcin-bazyl, line 101: I would also like to have reason added as an argument to setMuted() (with default value being MuteStateChangeReason.ByUser) so that we can use it for other things like remote mute - see https://github.com/webex/webex-js-sdk/blob/0692849ddb041ee642b9c2b6c589a54e7a242637/packages/%40webex/media-helpers/src/webrtc-core.ts#L29 for the current list of reasons used in SDK that we need on top of "by user" and "by browser"

src/media/remote-stream.ts
```diff
@@ -1,14 +1,87 @@
...
+   */
+  private addTrackHandlersForRemoteStreamEvents(track: MediaStreamTrack): void {
+    track.addEventListener('mute', this.handleMediaStopped);
+    track.addEventListener('unmute', this.handleMediaStarted);
+  }
+
+  /**
+   * @inheritdoc
+   */
+  protected addTrackHandlers(track: MediaStreamTrack): void {
```
marcin-bazyl, line 64: what happens when `RemoteStream.replaceTrack()` is called and the previous track was muted, but the new one isn't. Should we emit the MediaStateChange event?

bbaldino: There's no "mute" for remote (something we're trying to correct with new vocabulary), but I think this point still stands in that the media states could differ...maybe we need to compare the old and new states at replacement time.

bbaldino (dismissed):
lgtm, but let's figure out @marcin-bazyl 's question

src/media/remote-stream.ts
brycetham: That's a good point. Yeah, I think we should.

bbaldino (approved):

marcin-bazyl (approved):

PR #74, opened 2024-04-16
feat: expose on ice candidate event callback

k-wasniowski:
This PR exposes an additional event which will allow listening on generated ICE Candidates. 

`PeerConnection`:
* Add event `IceCandidate`, emits a `RTCPeerConnectionIceEvent` with gathered ICE Candidate Event

This change is provided to allow keeping track of generated ICE Candidates. Currently, it will be used to check if Firefox has properly generated relay candidate due to the existing bug -> https://bugzilla.mozilla.org/show_bug.cgi?id=1891503.

### DIFF
--- a/src/peer-connection.ts
+++ b/src/peer-connection.ts
@@ -27,6 +27,7 @@ type IceGatheringStateChangeEvent = {
 
 enum PeerConnectionEvents {
   IceGatheringStateChange = 'icegatheringstatechange',
+  IceCandidate = 'icecandidate',
   ConnectionStateChange = 'connectionstatechange',
   CreateOfferOnSuccess = 'createofferonsuccess',
   CreateAnswerOnSuccess = 'createansweronsuccess',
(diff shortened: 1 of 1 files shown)


k-wasniowski:
Not sure if it should expose `RTCPeerConnectionIceEvent`, or directly `RTCIceCandidate`.

brycetham (commented):
Looks good. Should we add some simple unit tests for this?

bbaldino:
Is this to enable a specific use case/scenario?  Don't necessarily disagree with it, but would be good to document the intent in the PR i think

bbaldino:
(I'd agree in general/for completeness it's a good idea though either way)

bbaldino (approved):

k-wasniowski:
This change is added to allow keeping track of generated ICE Candidates. Currently, it will be used to check if Firefox is properly generated relay candidate due to the existing bug -> https://bugzilla.mozilla.org/show_bug.cgi?id=1891503.

k-wasniowski:
> Looks good. Should we add some simple unit tests for this?

@brycetham I would like to, but not sure how such tests should look like. Do you have any proposal for such tests?

brycetham:
> > Looks good. Should we add some simple unit tests for this?
> 
> @brycetham I would like to, but not sure how such tests should look like. Do you have any proposal for such tests?

Honestly looking at it again, the only test I'd be interested is on in which it checks if `PeerConnection.Events.IceCandidate` gets emitted whenever `this.pc.onicecandidate` occurs. But the code is simple enough where I'm thinking it's probably not necessary.

brycetham (approved):

PR #78, opened 2024-06-10
SPARK-532493 webrtc core should expose separate events for ice and connection state

k-wasniowski:
This PR divide evaluating of overall connection state to two separated events.
We have decided that it's useful to have access to both events IceConnectionState and ConnectionState separately.
It's required for JS-SDK to have access directly to the reason of media join failure.


src/connection-state-handler.ts
```diff
@@ -113,6 +122,32 @@ export class ConnectionStateHandler extends EventEmitter<ConnectionStateEventHan
...
+   * Gets current ice connection state.
+   *
+   * @returns Current ice connection state.
+   */
+  public getIceConnectionState(): IceConnectionState {
+    const { iceState } = this.getCurrentStatesCallback();
+
+    const state = iceState[0].toUpperCase() + iceState.slice(1);
+
+    return state as IceConnectionState;
```
marcin-bazyl, line 142: I'm not sure why we need to define our own `IceConnectionState` enum. Wouldn't it be better to have `RTCIceConnectionState` as the return type for `getIceConnectionState()` and `RTCPeerConnectionState` as return type for `getConnectionState()`? Then we wouldn't have to be doing these tricks with converting first letter to uppercase etc. which I'm not a huge fan of....

src/connection-state-handler.ts
```diff
@@ -113,6 +122,32 @@ export class ConnectionStateHandler extends EventEmitter<ConnectionStateEventHan
...
+  }
+
+  /**
+   * Gets current overall connection state.
+   *
+   * @returns Current overall connection state.
+   */
+  public getOverallConnectionState(): OverallConnectionState {
```
brycetham, line 150: Wondering if it makes sense to keep calling this the `connectionState` and have the 2 sub-states be called `peerConnectionState` and `iceConnectionState`?

brycetham (commented):
Agreed with @marcin-bazyl's comments. Also left a comment about naming.

src/connection-state-handler.ts
k-wasniowski: Changed it as suggested. I like it, however had to change event name to keep it consistent. What do you think about it?

src/connection-state-handler.ts
k-wasniowski: Done

marcin-bazyl (dismissed):

src/connection-state-handler.spec.ts
```diff
@@ -17,39 +17,53 @@ describe('ConnectionStateHandler', () => {
...
+
+    expect(connStateHandler.getIceConnectionState()).toBe('new');
+  });
+
   it('reads initial connection state', () => {
     expect.assertions(1);
     const connStateHandler = new ConnectionStateHandler(fakeCallback);
 
-    expect(connStateHandler.getConnectionState()).toStrictEqual(ConnectionState.New);
+    expect(connStateHandler.getConnectionState()).toBe('New');
```
brycetham, line 38: Should this still be `ConnectionState.New`?

brycetham (approved):
Looks good, left a minor comment. As I discussed with you offline, since this is technically a breaking change, I would check with the SDK team to make sure they are ok with it, as this might affect some SDK customers. Likely they will need to document this somewhere, so I might wait until you get the ok before merging.

szotrabh (approved):

PR #81, opened 2024-06-27
feat: update typescript version

k-wasniowski:
This PR updates the typescript version. It's needed to be able to use directly `RTCPeerConnectionIceErrorEvent` which is not available in current typescript version. 

Typescript options `suppressImplicitAnyIndexErrors` is getting deprecated. It's removal causes not being able to compile because `suppressLocalAudioPlayback` is not defined for `MediaTrackConstraints`. 
`suppressLocalAudioPlayback` is experimental, thus it's not in typescript yet.

Are we safe to remove this flag from constraints or we should add an typescript option which will supress deprecation of `suppressImplicitAnyIndexErrors`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "semantic-release": "^19.0.2",
     "ts-jest": "^27.0.5",
     "ts-loader": "^9.2.3",
-    "typescript": "^4.4.3",
+    "typescript": "5.5.2",
     "yarn": "^1.22.18"
   },
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -72,7 +72,7 @@
     "semantic-release": "^19.0.2",
     "ts-jest": "^27.0.5",
     "ts-loader": "^9.2.3",
-    "typescript": "5.5.2",
+    "typescript": "^5.5.2",
     "yarn": "^1.22.18"
   },
   "scripts": {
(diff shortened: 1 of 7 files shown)


bbaldino:
I don't remember there being a specific reason for including `suppressLocalAudioPlayback`, I'm guessing it was copied over just as a result of including 'all' options that were possible there, so I think it should be ok to yank it?  cc @brycetham

k-wasniowski:
> I don't remember there being a specific reason for including `suppressLocalAudioPlayback`, I'm guessing it was copied over just as a result of including 'all' options that were possible there, so I think it should be ok to yank it? cc @brycetham

I did a quick search for it over JS-SDK/Web App and I have not found any usage of it.

brycetham:
Yeah I say we can get rid of `suppressLocalAudioPlayback`. We'll eventually want to allow any constraints to be applied instead of a predefined list anyways (see [WEBEX-386305](https://jira-eng-gpk2.cisco.com/jira/browse/WEBEX-386305)).

package.json
```diff
@@ -72,7 +72,7 @@
     "semantic-release": "^19.0.2",
     "ts-jest": "^27.0.5",
     "ts-loader": "^9.2.3",
-    "typescript": "^4.4.3",
+    "typescript": "5.5.2",
```
brycetham, line 75: Should we keep the `^`?

k-wasniowski: Added it back.

brycetham (approved):

PR #82, opened 2024-06-27
feat: expose icecandidateerror event

k-wasniowski:
[SPARK-536950](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-536950)

This PR introduces forwarding of IceCandidateError event. We would like to be able to use it in JS-SDK to have a metrics about most common ice gathering issues.

With typescript update to version > 5.0, cast would not be needed.

### DIFF
--- a/src/peer-connection.ts
+++ b/src/peer-connection.ts
@@ -28,6 +28,7 @@ type IceGatheringStateChangeEvent = {
 enum PeerConnectionEvents {
   IceGatheringStateChange = 'icegatheringstatechange',
   IceCandidate = 'icecandidate',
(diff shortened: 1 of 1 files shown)


bbaldino (approved):
lgtm, but maybe we want to wait for the typescript upgrade?

k-wasniowski:
> lgtm, but maybe we want to wait for the typescript upgrade?

Agree, created this PR just in case if we wouldn't want to upgrade typescript.

brycetham (dismissed):

brycetham (approved):

PR #86, opened 2024-09-19
fix: conditionally add device permissions

edvujic:
# Completes [SPARK-563488](https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-563488)

This PR introduces improvements and additions to the device management functionality, including conditional device permissions and unit tests. `getDevices()` function checks for all permissions irrespective of what the user has passed as `deviceKind`. This can cause unnecessary permission prompts for devices that aren't needed by the user.

This can be tested with: https://jira-eng-gpk2.cisco.com/jira/browse/SPARK-539683

### DIFF
--- a/src/device/device-management.spec.ts
+++ b/src/device/device-management.spec.ts
@@ -14,6 +14,7 @@ import {
   createMicrophoneStream,
   getDevices,
 } from './device-management';
+import { WebrtcCoreError, WebrtcCoreErrorType } from '../errors';
 
 jest.mock('../mocks/media-stream-stub');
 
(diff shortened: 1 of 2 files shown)


src/device/device-management.ts
```diff
@@ -190,11 +190,17 @@ export async function createDisplayStreamWithAudio<
...
 export async function getDevices(deviceKind?: media.DeviceKind): Promise<MediaDeviceInfo[]> {
   let devices: MediaDeviceInfo[];
+  const deviceKinds: media.DeviceKind[] = [];
+
+  // If deviceKind is provided, add it to the array. Otherwise, add both audio and video input kinds.
+  if (deviceKind !== undefined) {
+    deviceKinds.push(deviceKind);
+  } else {
+    deviceKinds.push(media.DeviceKind.AudioInput, media.DeviceKind.VideoInput);
+  }
```
bbaldino, line 200: Could this maybe be:

```typescript
const deviceKinds = deviceKind ? [deviceKind] : [media.DeviceKind.AudioInput, media.deviceKind.VideoInput];
```

bbaldino (approved):

src/device/device-management.spec.ts
```diff
@@ -254,4 +255,53 @@ describe('Device Management', () => {
       expect(localDisplayStream.contentHint).toBe('motion');
     });
   });
+
+  describe('getDevices', () => {
```
antsukanova, line 259: I'd also suggest covering cases with an error if we add `getDevices` tests. It's useful to understand how it should work, not only with the happy flow.

antsukanova (approved):
lgtm

PR #99, opened 2026-04-03
feat: re-acquire mic track when effects require specific audio constraints

antsukanova:
## Summary

Adds constraint handling to `LocalAudioStream` so audio effects can request specific `MediaTrackConstraints` on the microphone track. Since Chrome silently ignores `applyConstraints()` for audio processing properties ([chromium#40555809](https://issues.chromium.org/issues/40555809)), the handler re-acquires the track via `getUserMedia()`. Original track settings are saved and restored when the effect releases its constraints on disable/dispose.

Keeping the old track alive while calling `getUserMedia()` causes Chrome to reuse the existing hardware capture session and silently ignore hardware-level constraints such as `echoCancellation`. The old track is stopped **before** `getUserMedia()` to force a fresh session and ensure all constraints are honored.

## What's changed

- **`addConstraintHandlers`** — new private method on `LocalAudioStream` wired up from `addEffect()`; listens for `ConstraintsRequired` / `ConstraintsReleased` events from the effect and calls `reacquireInputTrack`
- **`reacquireInputTrack`** — stops the current track before calling `getUserMedia()` with the merged baseline + effect constraints; restores mute state and wires the new track into the effect chain
- **`AppliableAudioConstraints`** — extended with `sampleRate`, `sampleSize`, and `channelCount`
- **`filterToSupportedConstraints`** — new helper that strips keys not recognized by the browser via `getSupportedConstraints()` before passing baseline settings to `getUserMedia()`
- **`effects` / `loadingEffects` / `changeOutputTrack`** — visibility changed from `private` to `protected` in `LocalStream` to allow access from `LocalAudioStream`
- **Failure handling** — if `getUserMedia()` or effect wiring fails, all effects are disposed and `Ended` is emitted; if the effect is disposed while `getUserMedia()` is in-flight, the new track is discarded and `Ended` is emitted to prevent a silent dead stream

## Testing

- Enable OFMV → verify `autoGainControl: false`, `noiseSuppression: false`, `echoCancellation: true`
- Disable OFMV → verify original values restored
- BNR enable/disable → no constraint re-acquisition triggered
- Dispose during pending `getUserMedia()` → new track discarded, `Ended` emitted

See code change in [web-media-effects](https://github.com/beyonddunbar/web-media-effects/pull/228)  
You can use the sdk linked branch with these changes to test: [agc-fix-branch-test](https://github.com/antsukanova/webex-js-sdk/tree/antsukan/agc-fix-branch-test)


src/media/local-stream.ts
```diff
@@ -267,19 +268,110 @@ abstract class _LocalStream extends Stream {
...
+            if (!(key in savedConstraints)) {
+              savedConstraints[key] = oldSettings[key as keyof MediaTrackSettings];
+            }
+          });
+        }
+
+        this.removeTrackHandlers(oldTrack);
+        oldTrack.stop();
+
+        const newStream = await getUserMedia({
```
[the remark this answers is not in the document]
antsukanova: fixed with `this[StreamEventNames.Ended].emit();` and recovery init stream option

src/media/local-stream.ts
```diff
@@ -267,19 +268,129 @@ abstract class _LocalStream extends Stream {
       }
     };
 
+    /**
+     * Track settings saved before the effect changed them, keyed by constraint
+     * property name. Used to restore the user's original values when the effect
+     * emits empty constraints (disable / dispose / model switch to one with no
+     * special requirements).
+     */
+    let savedConstraints: Record<string, MediaTrackConstraints[keyof MediaTrackConstraints]> = {};
```
bbaldino: Do we have the possibility of multiple "layers" of saving constraints? Or currently can there only ever be one? I.e.: we have some constraints, a single thing can change them, and then the only other possible transition would be going back to those original constraints?

src/media/local-stream.ts
```diff
@@ -267,19 +268,129 @@ abstract class _LocalStream extends Stream {
...
+
+    /**
+     * Handle when an effect requests specific constraints on the input track.
+     *
+     * Non-empty constraints: save the current values for those properties, then
+     * re-acquire the mic track with the requested constraints.
+     *
+     * Empty constraints ({}): restore the previously saved values so the track
+     * returns to the user's original settings.
```
bbaldino: Out of curiosity: did you consider emitting different events for when new constraints are required and when they're no longer required (as opposed to overloading the value of the `constraints` parameter here)?  It seems like it would simplify the logic below a decent amount.

bbaldino (commented):
Couple comments.  Some impressive catches by codex as well 👀

src/media/local-stream.ts
antsukanova: Good comment, I thought about this too @bbaldino. The thing is, most of the handler logic like stopping the track, re-acquiring via `getUserMedia`, fallback, replacing the track — is the same whether we're applying or restoring constraints. So splitting into two events would just mean duplicating all of that or pulling it into a shared helper. Plus another event to wire up and test. Since this is only used for audio, it didn't feel worth the added complexity here.

src/media/local-stream.ts
antsukanova: For now it's always one `NoiseReductionEffect` instance with one active model at a time. I also updated the code logic so it's clear that this only applies to audio stream. I wouldn't want to introduce this for video as well since it's essentially a browser-specific workaround for `applyConstraints` being silently ignored.

The handler remembers the user's original settings before any model touched them, and restores everything on `disable()/dispose()`.

Maybe in the future this Chrome issue will be resolved and we can remove this layer entirely. I hope with my updates the logic is clearer, please take a look and let me know if you have any other comments. Thank you!

src/media/local-stream.ts
```diff
@@ -267,19 +268,136 @@ abstract class _LocalStream extends Stream {
...
+            },
+          });
+          [newTrack] = recoveryStream.getAudioTracks();
+          savedConstraints = {};
+        }
+
+        newTrack.enabled = wasEnabled;
+        this.inputStream.removeTrack(oldTrack);
+        this.inputStream.addTrack(newTrack);
+        this.addTrackHandlers(newTrack);
```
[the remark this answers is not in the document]
antsukanova: Fair point in theory, but I don't think this is a practical risk. `setModel()` already awaits worklet confirmation before emitting constraints, so model switches are naturally serialized. We'd need two emissions within a tiny window to actually race, which I can't see happening from normal user interaction.

src/media/local-stream.ts
```diff
@@ -267,20 +268,135 @@ abstract class _LocalStream extends Stream {
...
+     * Re-acquires via getUserMedia because MediaStreamTrack.applyConstraints()
+     * is silently ignored by Chrome for audio processing constraints.
+     * See https://issues.chromium.org/issues/40555809.
+     *
+     * @param constraints - The constraints requested by the effect.
+     */
+    const handleConstraintsRequired = async (constraints: MediaTrackConstraints) => {
+      if (!this.effects.includes(effect)) {
+        logger.log(`Effect ${effect.id} not in effects list, ignoring constraints-required.`);
+        return;
```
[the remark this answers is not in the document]
antsukanova: added new guard for this case.

src/media/local-stream.ts
bbaldino: Is it duplicating? Or is it just emitting the 'proper' event in each place, so the logic gets split between places that emit one event vs the other?

src/media/local-stream.ts
bbaldino: Ok cool, will take another pass at the latest changes.

src/media/local-stream.ts
```diff
@@ -292,6 +292,16 @@ abstract class _LocalStream extends Stream {
      * @param constraints - The constraints requested by the effect.
      */
     const handleConstraintsRequired = async (constraints: MediaTrackConstraints) => {
```
bbaldino: If this is now audio specific, should we rename it to indicate that as well?

antsukanova: Good point, can be as `handleAudioConstraintsRequired`
thank you for idea!

src/media/local-stream.ts
antsukanova: Fair point, the emission side wouldn't be duplication, you're right. The question is the handler side for me.

I asked AI to show this difference in practical way:
```ts
const handleAudioConstraintsRequired = async (constraints: MediaTrackConstraints) => {
  // --- guards (shared) ---
  if (!this.effects.includes(effect)) { return; }
  if (this.inputTrack.readyState === 'ended') { return; }

  // --- branching: 12 lines ---
  const isEmptyConstraints = !Object.keys(constraints).length;
  let constraintsToApply: MediaTrackConstraints;

  if (isEmptyConstraints) {
    if (!Object.keys(savedTrackSettings).length) { return; }
    constraintsToApply = { ...savedTrackSettings };
    savedTrackSettings = {};
  } else {
    constraintsToApply = constraints;
  }

  // --- shared re-acquisition: ~30 lines ---
  const currentTrack = this.inputTrack;
  const currentSettings = currentTrack.getSettings();
  const isEnabled = currentTrack.enabled;
  // ... isAlreadySatisfied check ...
  if (!isEmptyConstraints) { /* save settings */ }
  this.removeTrackHandlers(currentTrack);
  currentTrack.stop();
  // ... getUserMedia, fallback, replaceInputTrack ...
};
```

```ts
// --- shared helper: same ~30 lines extracted ---
const reacquireAudioTrack = async (constraintsToApply: MediaTrackConstraints) => {
  if (!this.effects.includes(effect)) { return; }
  if (this.inputTrack.readyState === 'ended') { return; }

  const currentTrack = this.inputTrack;
  const currentSettings = currentTrack.getSettings();
  const isEnabled = currentTrack.enabled;
  // ... isAlreadySatisfied check ...
  this.removeTrackHandlers(currentTrack);
  currentTrack.stop();
  // ... getUserMedia, fallback, replaceInputTrack ...
};

// --- handler 1: apply ---
const handleConstraintsApply = async (constraints: MediaTrackConstraints) => {
  Object.keys(constraints).forEach((key) => {
    if (!(key in savedTrackSettings)) {
      const currentSettings = this.inputTrack.getSettings();
      Object.assign(savedTrackSettings, { [key]: currentSettings[key as keyof MediaTrackSettings] });
    }
  });
  await reacquireAudioTrack(constraints);
};

// --- handler 2: restore ---
const handleConstraintsRestore = async () => {
  if (!Object.keys(savedTrackSettings).length) { return; }
  const toRestore = { ...savedTrackSettings };
  savedTrackSettings = {};
  await reacquireAudioTrack(toRestore);
};

// --- registration: 2 events instead of 1 ---
effect.on(EffectEvent.ConstraintsRequired, handleConstraintsApply);
effect.on(EffectEvent.ConstraintsRestored, handleConstraintsRestore);

// --- deregistration: 2 instead of 1 ---
effect.off(EffectEvent.ConstraintsRequired, handleConstraintsApply);
effect.off(EffectEvent.ConstraintsRestored, handleConstraintsRestore);
```

For me the split feels more scattered... three pieces (helper + two thin handlers) instead of one self-contained block. If this pattern were reused elsewhere, I'd agree it's worth the separation, but right now it's a single flow in one place, so the extra surface area (new event, extra registration/deregistration, more tests). So for this code, I don't have a strong opinion which feels better here, so I'm happy to discuss your view there.

src/media/local-stream.ts
```diff
@@ -267,20 +268,140 @@ abstract class _LocalStream extends Stream {
...
+
+        if (this.effects.length > 0) {
+          await this.effects[0].replaceInputTrack(newTrack);
+        }
+        this[LocalStreamEventNames.ConstraintsChange].emit();
+        logger.log(`Effect constraints applied via track re-acquisition.`);
+      } catch (err: unknown) {
+        logger.error(`Failed to re-acquire track after constraint change:`, err);
+        savedTrackSettings = {};
+        this[StreamEventNames.Ended].emit();
```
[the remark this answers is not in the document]
antsukanova: Catch emits `Ended` either way, which tells the SDK to set microphoneStream = undefined and discard the stream. Whether we stop the track or not, the stream is abandoned. And stopping a perfectly good mic track just to immediately tell the SDK "stream is dead" doesn't help the user — they have to re-create the stream regardless.

But i will think here to not emit `Ended` when only `replaceInputTrack` fails, maybe there could be better approach

src/media/local-stream.ts
```diff
@@ -267,19 +268,139 @@ abstract class _LocalStream extends Stream {
...
+        currentTrack.stop();
+
+        const deviceId = currentSettings.deviceId ? { exact: currentSettings.deviceId } : undefined;
+
+        let newStream = await getUserMedia({
+          audio: { ...currentSettings, ...constraintsToApply, deviceId },
+        }).catch((err) => {
+          logger.warn(`Failed to re-acquire track with effect constraints, recovering:`, err);
+          return null;
+        });
```
k-wasniowski: Do we need to handle this catch here? Or it would be fine to let it flow to upper layer try/catch?

k-wasniowski: Actually, nvm

edvujic (dismissed):

src/media/local-stream.ts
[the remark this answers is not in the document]
antsukanova: From my checks best way seems to do here:

Split the catch block by checking two flows:

1. if `this.inputTrack.readyState` is 'live': fall back to the raw mic track via `changeOutputTrack(this.inputTrack)`, which routes it to the sender through the existing `OutputTrackChange` event. No stream-ended, audio continues.
2. getUserMedia failed, no mic: emit stream-ended as before.

src/media/local-stream.ts
antsukanova: This is here to make sure that if a small setting like `{ autoGainControl: false }` doesn’t work on a certain device or browser, the audio stream won’t stop. Instead, it falls back to the user’s original settings, which were working before. Should be super rare, but just in case, it would be better to have a fallback here

edvujic (dismissed):

src/media/local-stream.ts
bbaldino: I don't know if I love the naming (`handleConstraintsApply` and `handleConstraintsRestore`) but I do think it's nice that we end up with more focused events/methods and the handling is cleaner.  I'll defer to you on the final call though.

src/media/local-stream.ts
```diff
@@ -291,7 +291,7 @@ abstract class _LocalStream extends Stream {
      *
      * @param constraints - The constraints requested by the effect.
      */
-    const handleConstraintsRequired = async (constraints: MediaTrackConstraints) => {
+    const handleAudioConstraintsRequired = async (constraints: MediaTrackConstraints) => {
```
bbaldino: Reading this again, I wonder if we can do a better name than `handleAudioConstraintsRequired`.  It's not _just_ that constraints are required here, it's that we need _additional_ constraints added, maybe?  Is it more like 'handleAdditionalAudioConstraints`, maybe?

src/media/local-stream.ts
```diff
@@ -267,19 +268,154 @@ abstract class _LocalStream extends Stream {
...
+     * Re-acquires via getUserMedia because MediaStreamTrack.applyConstraints()
+     * is silently ignored by Chrome for audio processing constraints.
+     * See https://issues.chromium.org/issues/40555809.
+     *
+     * @param constraints - The constraints requested by the effect.
+     */
+    const handleAudioConstraintsRequired = async (constraints: MediaTrackConstraints) => {
+      if (!this.effects.includes(effect)) {
+        logger.log(`Effect ${effect.id} not in effects list, ignoring constraints-required.`);
+        return;
```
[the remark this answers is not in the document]
antsukanova: When a same-kind effect is replaced:

1. splice removes the old effect from this.effects and puts the new one in
2. New effect enable() runs — its constraints take effect
3. Old effect dispose() runs — emits ConstraintsReleased

But the old effect's `ConstraintsReleased` is harmless because `reacquireInputTrack` checks this.effects.includes(effect) at the top. The old effect was already spliced out, so the check fails and the event is ignored. The new effect's constraints stay in place.

With new logic with `constraintsOverridden` in media effects should be not an issue.

src/media/local-stream.ts
```diff
@@ -267,20 +268,170 @@ abstract class _LocalStream extends Stream {
...
+
+        const [newTrack] = newStream.getAudioTracks();
+
+        // If the effect was disposed while getUserMedia was pending, stop the
+        // newly acquired track and bail out to avoid reinserting a live mic
+        // track into a stopped stream (same pattern as addEffect).
+        if (!this.effects.includes(effect)) {
+          newTrack.stop();
+          logger.log(`Effect was disposed during track re-acquisition, discarding new track.`);
+          return;
```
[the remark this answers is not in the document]
antsukanova: Good concern, updated with deferring logic for `currentTrack.stop()` until after `getUserMedia` succeeds.

src/media/local-stream.ts
```diff
@@ -267,19 +268,139 @@ abstract class _LocalStream extends Stream {
...
+          });
+          savedTrackSettings = {};
+        }
+
+        const [newTrack] = newStream.getAudioTracks();
+
+        newTrack.enabled = isEnabled;
+        this.inputStream.removeTrack(currentTrack);
+        this.inputStream.addTrack(newTrack);
+        this.addTrackHandlers(newTrack);
```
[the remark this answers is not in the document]
antsukanova: It’s a good one. I have duplicated the flow from the start of the function checks to address this issue.

src/media/local-stream.ts
antsukanova: Thanks for idea, but I think `handleAudioConstraintsRequired` is the right name. It's an event handler for the `ConstraintsRequired` event, so handle + event name keeps the mapping clear and stays symmetric with `handleAudioConstraintsReleased` as well. The effect isn't asking to add constraints on top of existing ones it's saying "I require these specific values to function correctly." The handler then overrides the track settings to satisfy them.
Does it still not seem good?

antsukanova:
@bbaldino Thanks for your review! I have updated PR with two events, and I think sticking with `handleAudioConstraintsRequired` / `handleAudioConstraintsReleased` is the better fit! They map directly to the event names. If you have a chance to check, I would much appreciate your view here

src/media/local-stream.ts
```diff
@@ -267,20 +268,172 @@ abstract class _LocalStream extends Stream {
...
+    const handleAudioConstraintsRequired = async (
+      constraints: MediaTrackConstraints
+    ): Promise<void> => {
+      logger.log(`Effect ${effect.id} constraints required:`, constraints);
+
+      const currentSettings = this.inputTrack.getSettings();
+      Object.keys(constraints).forEach((key) => {
+        if (!(key in savedTrackSettings)) {
+          Object.assign(savedTrackSettings, {
+            [key]: currentSettings[key as keyof MediaTrackSettings],
```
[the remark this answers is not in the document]
antsukanova: All consumers follow the same pattern: call `getEffectByKind()` first, only `addEffect` if nothing exists yet, then toggle `enable()/disable()` on that single instance.

The `savedTrackSettings` snapshot concern only applies if `addEffect` is called with a new instance while an old one of the same kind is still active which no real consumer does today.

k-wasniowski (dismissed):

edvujic (dismissed):

src/media/local-stream.ts
```diff
@@ -267,20 +268,172 @@ abstract class _LocalStream extends Stream {
...
      * effect.
      */
     const handleEffectDisposed = () => {
-      effect.off('track-updated' as EffectEvent, handleEffectTrackUpdated);
-      effect.off('disposed' as EffectEvent, handleEffectDisposed);
+      effect.off('track-updated' as EffectEvent, handleEffectTrackUpdated as never);
+      if (this.outputTrack.kind === 'audio') {
+        effect.off('constraints-required' as EffectEvent, handleAudioConstraintsRequired as never);
+        effect.off('constraints-released' as EffectEvent, handleAudioConstraintsReleased as never);
+      }
```
bbaldino: we have `local-audio-stream`, don't we?  If so, could we have this audio-specific logic (and maybe the stuff above too) done there?

antsukanova: Good point @bbaldino. I considered this as well. My initial concern was that moving it to `LocalAudioStream` would require changing some fields from `private` to `protected`, which felt like an API change driven by a workaround. But thinking more about it, keeping this logic in `LocalStream` is not so great as well because it affects all stream types when it should be an audio-only internal detail. The protected change is a reasonable trade-off since it only exposes `effects` and `changeOutputTrack` to subclasses.

I've updated the code to move the constraint handling into a `private` `addConstraintHandlers` method on `LocalAudioStream`. Could you take a look and let me know if the approach seems better for you as well? Thanks!

src/media/local-audio-stream.ts
```diff
@@ -11,6 +14,14 @@ export type AppliableAudioConstraints = Pick<
  * An audio LocalStream.
  */
 export class LocalAudioStream extends LocalStream {
+  /**
+   * @inheritdoc
+   */
+  async addEffect(effect: TrackEffect): Promise<void> {
+    await super.addEffect(effect);
+    this.addConstraintHandlers(effect);
```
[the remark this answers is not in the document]
antsukanova: With current consumers, both orderings produce identical behavior, but it costs nothing to swap so can be done just being defensively correct.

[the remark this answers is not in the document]
antsukanova: Placing handlers after a`wait super.addEffect(effect)` ensures they're only registered when the effect is successfully added — no dangling listeners on load failures.

The trade-off is that same-kind replacement (where enable() emits `ConstraintsRequired` before handlers are attached) isn't covered. But no current consumer calls `addEffect` without disposing the old effect first, so this path is never hit. Safe to keep as-is for now.

src/media/local-audio-stream.ts
```diff
@@ -11,6 +14,17 @@ export type AppliableAudioConstraints = Pick<
...
 export class LocalAudioStream extends LocalStream {
+  /**
+   * @inheritdoc
+   */
+  async addEffect(effect: TrackEffect): Promise<void> {
+    if (this.effects.some((e) => e.id === effect.id)) {
+      return;
+    }
+    this.addConstraintHandlers(effect);
+    await super.addEffect(effect);
```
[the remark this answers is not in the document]
antsukanova: https://github.com/webex/webrtc-core/pull/99#discussion_r3094929461

src/media/local-audio-stream.ts
```diff
@@ -23,4 +37,165 @@ export class LocalAudioStream extends LocalStream {
...
+    let savedTrackSettings: MediaTrackSettings = {};
+
+    /**
+     * Re-acquire the audio input track with the given constraints via
+     * getUserMedia, since Chrome ignores applyConstraints for audio
+     * processing properties.
+     *
+     * @param constraintsToApply - The constraints to apply to the new track.
+     */
+    const reacquireInputTrack = async (
```
k-wasniowski, line 62: @antsukanova Would it be possible for this method to be called twice in a row? User switching selected effect rapidly, switching camera or any other scenario? I believe in such case calling this method twice in a row would be very bad race condition

src/media/local-audio-stream.ts
```diff
@@ -23,4 +37,165 @@ export class LocalAudioStream extends LocalStream {
...
+        return;
+      }
+
+      if (this.inputTrack.readyState === 'ended') {
+        logger.log(`Track already ended, ignoring constraints change.`);
+        return;
+      }
+
+      const currentTrack = this.inputTrack;
+      const currentSettings = currentTrack.getSettings();
```
k-wasniowski, line 76: Not sure if it matter. We are getting here settings from current track which has type `MediaTrackSettings`. Then re-use to create from it `MediaTrackConstraints`. I see that most of the fields are the same.
But for example `MediaTrackConstraints` doesnt have [restrictOwnAudio](https://developer.mozilla.org/en-US/docs/Web/API/MediaTrackSettings/restrictOwnAudio) which exists for  settings.
It might end up raising an [OverconstrainedError](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia) from some browsers.

I assume that most will ignore params which they do not understand.

src/media/local-audio-stream.ts
```diff
@@ -23,4 +37,165 @@ export class LocalAudioStream extends LocalStream {
...
+      );
+      if (isAlreadySatisfied) {
+        logger.log(`Constraints already satisfied, skipping re-acquisition.`);
+        return;
+      }
+
+      try {
+        const deviceId = currentSettings.deviceId ? { exact: currentSettings.deviceId } : undefined;
+
+        let newStream = await getUserMedia({
```
k-wasniowski, line 90: If we fail to apply new constraints, should we emit some event?

k-wasniowski (approved):

src/media/local-audio-stream.ts
antsukanova: Good question! 
Should be fine due to:
1. Unknown property names are silently dropped (no OverconstrainedError) for ssuch properties
https://www.w3.org/TR/mediacapture-streams/  (11. Constrainable Pattern)

but as option we can have `getSupportedConstraints` check for preserved props to filter them only with valid ones before woring with them, should be fine to add I assume

src/media/local-audio-stream.ts
antsukanova: In practice this is unlikely because `ConstraintsReleased` only will be called after the async worklet disable completes, by which time the first getUserMedia has usually I believe already resolved
I believe user would need to toggle the button in UI with a super fast speed to really have this issue
Video is not problem there just in case

Upd: for sdk it is possible but I do not feel in web client will be have this issue because it's not a one button to switch states

src/media/local-audio-stream.ts
antsukanova: Good question. I looked at this and I don't think we need a new event here.
The two events we already emit cover both real outcomes:
- `ConstraintsChange` — fires whenever a new track is wired in, whether it ended up with the effect's requested constraints or fell back to the original ones.
- `Ended` — fires if both `getUserMedia` calls fail and the track is gone, which is exactly when the consumer needs to tell the server "no audio source anymore."

I feel adding a new one would expand the public API for a Chrome workaround we want to delete cleanly it later.

src/media/local-audio-stream.ts
```diff
@@ -1,26 +1,227 @@
...
+        if (this.inputTrack.readyState === 'live') {
+          // Both getUserMedia attempts failed but the original mic is still alive.
+          // Bypass the effect and send audio straight from the mic.
+          logger.error(`Effect wiring failed, disposing effect and continuing with raw mic:`, err);
+          this.changeOutputTrack(this.inputTrack);
+          const index = this.effects.indexOf(effect);
+          if (index >= 0) {
+            this.effects.splice(index, 1);
+          }
+          await effect.dispose().catch((disposeErr) => {
```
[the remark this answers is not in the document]
antsukanova: actually good catch, i have updated with disposeEffects api

src/media/local-audio-stream.ts
```diff
@@ -1,26 +1,228 @@
...
+
+        if (this.inputTrack.readyState === 'live') {
+          // Both getUserMedia attempts failed but the original mic is still alive.
+          // Bypass the effect and send audio straight from the mic.
+          logger.error(`Effect wiring failed, disposing effect and continuing with raw mic:`, err);
+          this.changeOutputTrack(this.inputTrack);
+          const index = this.effects.indexOf(effect);
+          if (index >= 0) {
+            this.effects.splice(index, 1);
+          }
```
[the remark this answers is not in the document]
antsukanova: https://github.com/webex/webrtc-core/pull/99#discussion_r3110618874

src/media/local-audio-stream.ts
```diff
@@ -1,26 +1,223 @@
...
+          logger.log(`Effect was disposed during constraint handling, ignoring error.`);
+          return;
+        }
+
+        if (this.inputTrack.readyState === 'live') {
+          // Mic is still live but the effect chain is broken. Tear down all
+          // effects so the raw mic plays through and getEffects() matches the
+          // actual audio path.
+          logger.error(`Effect wiring failed, falling back to raw mic:`, err);
+          await this.disposeEffects().catch((disposeErr) => {
```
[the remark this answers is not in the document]
antsukanova: Replaced `disposeEffects()` with an inline dispose to not complicate code too much with additional states

src/media/local-audio-stream.ts
```diff
@@ -1,26 +1,241 @@
...
+        this.removeTrackHandlers(currentTrack);
+        currentTrack.stop();
+
+        // Preserve the mute state across the track swap.
+        newTrack.enabled = currentTrack.enabled;
+        this.inputStream.removeTrack(currentTrack);
+        this.inputStream.addTrack(newTrack);
+        this.addTrackHandlers(newTrack);
+
+        await effect.replaceInputTrack(newTrack);
```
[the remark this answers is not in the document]
antsukanova: In our current product path, audio is effectively single-effect (noise reduction only), so this doesn’t impact real usage today

src/media/local-audio-stream.ts
```diff
@@ -1,26 +1,241 @@
+import { EffectEvent } from '@webex/web-media-effects';
+import { getUserMedia } from '.';
 import { logger } from '../util/logger';
-import { LocalStream, LocalStreamEventNames } from './local-stream';
+import { StreamEventNames } from './stream';
+import { LocalStream, LocalStreamEventNames, TrackEffect } from './local-stream';
 
-// These are the audio constraints that can be applied via applyConstraints.
+// Subset of audio constraints that can be applied to a live track.
```
bbaldino: What's the reason for this distinction? I think originally this was intended to model _all_ audio-specific constraints (i.e. the ones that can be used in `getUserMedia`).  If we want to model the ones that can only be applied to live tracks, maybe we should have a different type for that?

src/media/local-audio-stream.ts
```diff
@@ -1,26 +1,241 @@
...
+        }).catch((err) => {
+          logger.warn(`Failed to re-acquire track with effect constraints, recovering:`, err);
+          return null;
+        });
+
+        if (!newStream) {
+          newStream = await getUserMedia({
+            audio: { ...baselineConstraints, deviceId },
+          });
+        }
```
bbaldino: In what scenario is this fallback hit?

src/media/local-audio-stream.ts
```diff
@@ -1,26 +1,241 @@
...
+
+        const [newTrack] = newStream.getAudioTracks();
+        if (!newTrack) {
+          logger.warn(`Re-acquire returned no audio track, skipping replacement.`);
+          return false;
+        }
+
+        // The effect may have been removed or the track may have ended while
```
bbaldino: If this effect is removed, that will fire another event, right?

I'm wondering if this is worth adding.  If it _is_, then we could technically be doing this check after _every_ await point (since the effect could've been removed at any of those suspension points).

src/media/local-audio-stream.ts
```diff
@@ -1,26 +1,241 @@
...
+        // getUserMedia was running. Discard the new track so it doesn't keep
```
bbaldino: I don't follow what's meant by "getUserMedia was running" here?

src/media/local-audio-stream.ts
```diff
@@ -1,26 +1,241 @@
...
+        // the microphone open in the background.
+        if (!this.effects.includes(effect) || currentTrack.readyState === 'ended') {
+          newTrack.stop();
+          logger.log(`Effect was disposed during track re-acquisition, discarding new track.`);
+          return false;
+        }
```
bbaldino: This is another re-check of the logic we have at the start of the method, right? Wondering how paranoid we need to be.

src/media/local-audio-stream.ts
```diff
@@ -1,26 +1,241 @@
...
+
+        this.removeTrackHandlers(currentTrack);
+        currentTrack.stop();
+
+        // Preserve the mute state across the track swap.
+        newTrack.enabled = currentTrack.enabled;
+        this.inputStream.removeTrack(currentTrack);
+        this.inputStream.addTrack(newTrack);
+        this.addTrackHandlers(newTrack);
```
bbaldino, line 145: Do we need a corresponding 'removeTrackHandlers' call for currentTrack?

src/media/local-audio-stream.ts
```diff
@@ -1,26 +1,241 @@
...
+
+        if (this.inputTrack.readyState === 'live') {
+          // Mic is still live but the effect chain is broken — fall back to raw audio.
+          // Clear this.effects before disposing so any constraints-released events
+          // fired from inside dispose() are skipped by the reacquire guards.
+          logger.error(`Effect wiring failed, falling back to raw mic:`, err);
+          this.changeOutputTrack(this.inputTrack);
+          // If another effect is still loading, drop it too so it can't
+          // slip back into the chain once we've fallen back to raw mic.
+          this.loadingEffects.clear();
```
bbaldino: Could this cause a problem when multiple effects are added/loading? Might there be a case where we don't want to impact another effect?

bbaldino (commented):
This is tough...the logic is getting pretty complex here.  Left some some comments.

src/media/local-audio-stream.ts
antsukanova: Good catch. The "live track" wording was likely added because this area was touched as part of the Chrome applyConstraints bug — but you're right, it's misleading. All `MediaTrackConstraints` can technically be passed to `applyConstraints`. 

Updated the comment and the type to match video stream pattern.

src/media/local-audio-stream.ts
antsukanova: @bbaldino The only idea behind it was: if the browser rejects the effect's constraint values specifically (e.g. they're not supported on that hardware) but would still open the mic fine with the baseline settings alone. In practice I've never hit this and looking at the docs it seems highly unlikely...Chrome silently accepts these flags even when it can't apply them (that's literally the bug this PR fixes), and for any real failure the catch block already handles recovery correctly. Happy to simplify and let catch be the single failure path. Thanks for flagging.

src/media/local-audio-stream.ts
antsukanova: Yes, removing the effect fires disposed which removes our listeners — but that only stops future events from firing. It doesn't stop a handler that's already mid-execution waiting on getUserMedia. 

One guard here is enough because getUserMedia is the only await where real time passes.
But believe we need to keep it due to my rapid enable/disable race flow: https://github.com/webex/webrtc-core/pull/99/changes/BASE..908a40b6feaf71a422155872f3e356fe341b7c64#r3208557639

src/media/local-audio-stream.ts
antsukanova: Got your confusion, then suggest here a bit clearer comment about `"While we were awaiting getUserMedia"`, thanks

src/media/local-audio-stream.ts
antsukanova: This is the only guard that catches the rapid enable/disable race, both calls pass the entry guard because at the moment each starts the effect is still there and the track is still live. Only after the first call completes and stops `currentTrack` does the state change, and by then the second call is already past the entry guard waiting on `getUserMedia`. So I would consider keeping it for even that reason, I had that in the SDK smaple app

src/media/local-audio-stream.ts
antsukanova: `removeTrackHandlers` is already there on line 134~, called before `currentTrack.stop()` — which one are you referring to?

src/media/local-audio-stream.ts
antsukanova: In practice `LocalAudioStream` today only uses noise reduction effects and they don't chain together. But even if a second effect were loading, clearing it here is the right call I think. Effects depend on each other's output track, and we've just torn down the chain and switched to raw mic. A loading effect that completes after this would try to wire itself to a track that no longer exists. Dropping it and letting the caller retry is cleaner than letting a half-loaded effect slip into a broken chain. But again now it is only one effect per audio

src/media/local-audio-stream.ts
```diff
@@ -1,26 +1,241 @@
...
+          logger.log(`Effect was disposed during constraint handling, ignoring error.`);
+          return false;
+        }
+
+        if (this.inputTrack.readyState === 'live') {
+          // Mic is still live but the effect chain is broken — fall back to raw audio.
+          // Clear this.effects before disposing so any constraints-released events
+          // fired from inside dispose() are skipped by the reacquire guards.
+          logger.error(`Effect wiring failed, falling back to raw mic:`, err);
+          this.changeOutputTrack(this.inputTrack);
```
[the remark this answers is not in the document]
antsukanova: I believe this is really rare, but it is not a big change, so will add it, but it is discussible.

src/media/local-audio-stream.ts
```diff
@@ -1,26 +1,251 @@
...
+        }
+
+        // Preserve the mute state before the effect receives the new track.
+        newTrack.enabled = currentTrack.enabled;
+
+        // Try wiring the effect before committing the stream swap. If this
+        // fails, close the unused new track and let the outer catch fall back
+        // to the still-live current track.
+        try {
+          await effect.replaceInputTrack(newTrack);
```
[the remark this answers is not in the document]
antsukanova: The existing post `getUserMedia()` guard covers the practical slow race, I would not like coplicate this here and add new `if`.

src/media/local-audio-stream.ts
bbaldino: Ok, yeah given how complex all the flows are it would be able to yank this if we could do so safely.

src/media/local-audio-stream.ts
bbaldino: There are other await points here though, right? (Can't easily check it now).  Any await/suspension point could happen for any duration of time, we can't know for sure.

src/media/local-audio-stream.ts
bbaldino: >  A loading effect that completes after this would try to wire itself to a track that no longer exists.  Dropping it and letting the caller retry 

would the caller be aware that this had happened though?

src/media/local-audio-stream.ts
bbaldino: Should we look at serializing these calls explicitly, like we do with sld/srd? like streams could have a serialization queue that some operations need to pass through so we could minimize awkward races like this?

src/media/local-audio-stream.ts
```diff
@@ -1,26 +1,231 @@
...
   /**
-   * Apply constraints to the stream.
+   * @inheritdoc
+   */
+  async addEffect(effect: TrackEffect): Promise<void> {
+    if (this.effects.some((e) => e.id === effect.id)) {
+      return;
+    }
+    await super.addEffect(effect);
+    this.addConstraintHandlers(effect);
```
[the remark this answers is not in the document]
antsukanova: this should not exist in practice with sdk usage

src/media/local-audio-stream.ts
antsukanova: I have a bit changes in logic here and now it has simplified flow, so we have only one `if (!this.effects.includes(effect)) {` just after `await getUserMedia({` to check if effect was disposed during `getUserMedia`. 

We have other awaits here, but they feel ok to skip this check and not overcomplicate. Ideally, I wish to remove this workaround altogether if we fix `applyContrains`.

src/media/local-audio-stream.ts
antsukanova: Good analogy, thank you for mentioning that! But the two cases are seems a bit different in terms of impact. With SLD/SRD, I believe race breaks the PC in a way you can't recover from. Here the worst that happens is the mic keeps slightly wrong settings until the user toggles again, and it fixes itself automatically (but webex-web-client handles that already).

Also I have checked webex-web-client and it already blocks rapid switches at the Redux level, so regular users can't actually hit this. SDK developers building their own UI could technically trigger it, but they'd need to click within a ~200ms window while the mic is reopening.

For now I'd lean towards keeping it simple and skipping the queue, but genuinely open to your take if you feel it's worth adding, happy to discuss.

src/media/local-audio-stream.ts
antsukanova: Yes. The caller gets notified through two signals: the addEffect promise rejects with an `ADD_EFFECT_FAILED` error (the base class detects the cleared loading map and throws), and the stream also emits Ended. So nothing happens silently here, the caller can catch the rejection and react, or listen for Ended to know the stream is gone.

But this path is actually theoretical for audio today. Nobody calls `addEffect` on an audio stream while another effect is already loading.

k-wasniowski (dismissed):

fnowakow (approved):

szotrabh (dismissed):

antsukanova:
Updated with latest `main` and resolved the package/lockfile conflicts. I also addressed the three current correctness findings: preserve mute changes while `replaceInputTrack()` is pending, emit `OutputTrackChange` when the raw public track is replaced, and emit `Ended` if a stale re-acquisition rejects after the old mic was stopped.

Local verification on Node 22/Yarn 1:
- `yarn build` ✅
- `yarn test:lint` ✅
- `yarn test:prettier` ✅
- `yarn test:spelling` ✅
- focused `local-audio-stream.spec.ts`: 20/20 ✅
- `yarn test:coverage`: 154/154 ✅
- no credential, key, `.pem`, certificate, or `.env` files staged ✅

CI is rerunning. @bbaldino @fnowakow, please provide the final approval when convenient.

antsukanova:
Follow-up: I reverted the three additional correctness changes from the previous update because I want to verify them first and confirm myself

src/media/local-microphone-stream.ts
```diff
@@ -1,6 +1,203 @@
...
+      const currentSettings = this.inputTrack.getSettings();
+      /**
+       * Save a single setting if not already saved.
+       *
+       * @param key - The setting key to save.
+       */
+      const snapshot = <K extends keyof MediaTrackSettings>(key: K): void => {
+        if (!(key in savedTrackSettings) && key in currentSettings) {
+          savedTrackSettings[key] = currentSettings[key];
+        }
```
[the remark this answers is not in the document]
antsukanova: I checked the current full flow: only `NoiseReductionEffect` emits these events, and OFMV/BNR are models of the same effect kind, so only one constraint-owning effect exists on the microphone stream. The scenario is therefore not reachable today.

src/media/local-microphone-stream.ts
```diff
@@ -1,6 +1,203 @@
...
+      };
+      (Object.keys(constraints) as Array<keyof MediaTrackSettings>).forEach(snapshot);
```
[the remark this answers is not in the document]
antsukanova: No current path emits advanced constraints and `NoiseReductionEffect.getRequiredConstraints()` emits only flat boolean APM constraints: `autoGainControl`, `noiseSuppression`, and `echoCancellation`.

src/media/local-audio-stream.ts
```diff
@@ -1,26 +1,246 @@
...
+      const deviceId = currentSettings.deviceId ? { exact: currentSettings.deviceId } : undefined;
+      const baselineConstraints = filterToSupportedConstraints(currentSettings);
+
+      try {
+        this.removeTrackHandlers(currentTrack);
+        currentTrack.stop();
+
+        const newStream = await getUserMedia({
+          audio: { ...baselineConstraints, ...constraintsToApply, deviceId },
+        });
```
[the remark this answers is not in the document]
antsukanova: Valid: addressed this by moving the complete constraint re-acquisition workaround from `LocalAudioStream` to `LocalMicrophoneStream`

src/media/local-audio-stream.ts
```diff
@@ -1,26 +1,232 @@
...
+          newTrack.stop();
+          logger.log(`Effect was disposed during getUserMedia, emitting Ended.`);
+          this[StreamEventNames.Ended].emit();
+          return false;
+        }
+
+        newTrack.enabled = currentTrack.enabled;
+
+        try {
+          await effect.replaceInputTrack(newTrack);
```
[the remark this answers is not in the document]
antsukanova: implemented

src/media/local-audio-stream.ts
```diff
@@ -1,26 +1,232 @@
...
+
+        this[LocalStreamEventNames.ConstraintsChange].emit();
+        logger.log(
+          `Constraints applied via track re-acquisition. Settings:`,
+          newTrack.getSettings()
+        );
+        return true;
+      } catch (err: unknown) {
+        if (!this.effects.includes(effect)) {
+          return false;
```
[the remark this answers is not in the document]
antsukanova: it is extremely unlikely to happen in current implementation.

src/media/local-microphone-stream.ts
```diff
@@ -1,6 +1,206 @@
...
+      constraintsToApply: MediaTrackConstraints
+    ): Promise<boolean> => {
+      if (!this.effects.includes(effect)) {
+        logger.log(`Effect ${effect.id} is no longer active, skipping constraint handling.`);
+        return false;
+      }
+
+      if (this.inputTrack.readyState === 'ended') {
+        logger.log(`Track already ended, ignoring constraints change.`);
+        return false;
```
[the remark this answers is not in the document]
antsukanova: improved

src/media/local-stream.ts
```diff
@@ -377,11 +377,14 @@ abstract class _LocalStream extends Stream {
   async disposeEffects(): Promise<void> {
     this.loadingEffects.clear();
 
-    // Dispose of any effects currently in use
-    if (this.effects.length > 0) {
+    // Effects are no longer active once disposal begins. Clear them before invoking
+    // disposal callbacks so those callbacks cannot start new work on this stream.
+    const effectsToDispose = this.effects;
```
antsukanova: change for `ConstraintsReleased` correct work

src/media/local-microphone-stream.ts
```diff
@@ -1,6 +1,206 @@
...
+
+        const [newTrack] = newStream.getAudioTracks();
+        if (!newTrack) {
+          throw new Error('getUserMedia returned a stream with no audio tracks.');
+        }
+
+        if (!this.effects.includes(effect)) {
+          newTrack.stop();
+          logger.log(`Effect was disposed during getUserMedia, emitting Ended.`);
+          this[StreamEventNames.Ended].emit();
```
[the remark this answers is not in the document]
antsukanova: current product flow, it is too rare to handle now

src/media/local-microphone-stream.ts
```diff
@@ -1,6 +1,206 @@
...
+      try {
+        this.removeTrackHandlers(currentTrack);
+        currentTrack.stop();
+
+        const newStream = await getUserMedia({
+          audio: { ...baselineConstraints, ...constraintsToApply, deviceId },
+        });
+
+        const [newTrack] = newStream.getAudioTracks();
+        if (!newTrack) {
```
BlueLamantine: If `getUserMedia` returns a stream with no audio tracks, should we stop any leftover tracks on that stream before throwing?

src/media/local-microphone-stream.ts
```diff
@@ -1,6 +1,206 @@
...
+     */
+    const reacquireInputTrack = async (
+      constraintsToApply: MediaTrackConstraints
+    ): Promise<boolean> => {
+      if (!this.effects.includes(effect)) {
+        logger.log(`Effect ${effect.id} is no longer active, skipping constraint handling.`);
+        return false;
+      }
+
+      if (this.inputTrack.readyState === 'ended') {
```
BlueLamantine: If OFMV is released while `getUserMedia` is still pending, `handleConstraintsReleased` sees `readyState === 'ended'` and bails. The in-flight acquire can then apply OFMV constraints after disable. Should we serialize these transitions?

src/media/local-microphone-stream.ts
```diff
@@ -1,6 +1,206 @@
...
+          newTrack.stop();
+          logger.log(`Effect was disposed during getUserMedia, emitting Ended.`);
+          this[StreamEventNames.Ended].emit();
+          return false;
+        }
+
+        newTrack.enabled = currentTrack.enabled;
+
+        try {
+          await effect.replaceInputTrack(newTrack);
```
BlueLamantine: `replaceInputTrack` is called on the emitting effect, not `this.effects[0]`. If another effect sits in front of OFMV, does it stay wired to the stopped track? Should we replace the head of the chain and let track-updated propagate?

src/media/local-microphone-stream.ts
antsukanova: good question. In the current sdk/web client, we always use `getUserMedia()` with audio-only constraints. In this case, the browser should either return an audio track or reject the request. I would say here we should be safe for now to not add this and do not complicate if we not expecting this, just logging for analys.

src/media/local-microphone-stream.ts
antsukanova: Good question. For the current Web Client flow, I think this risk is negligible. The Smart Audio menu is disabled while the action is pending, so the user cannot immediately disable OFMV.

The race requires getUserMedia() to remain pending after the menu becomes available, followed by the user immediately disabling OFMV before capture completes. 

Adding a queue would increase complexity based on a theoretical case. I would suggest to keep that in mind on web client development stage and to validate the timing then and adding serialization only if we can reproduce it.

src/media/local-microphone-stream.ts
antsukanova: Good question! In the current Web Client and SDK flow, `NoiseReductionEffect` is the only microphone effect, so the requesting effect is also `this.effects[0]`. Changing the call to the first effect would not affect the current flow and would not fully solve this question easly with chain rewiring because every effect must ideally reconnect as well then. 
I'd suggest keeping this PR scoped to the current OFMV flow and handling multi-effect flow separately when we have a real case for it. (maybe this fix will be not needed before this issue arise)

BlueLamantine (approved):

PR #100, opened 2026-05-12
feat: peer connection support setConfiguration api

k-wasniowski:
## Description

Expose setConfiguration function with peerconnection API

## This change implements...
- [x] A new feature
- [ ] A bug fix
- [ ] Other (please specify):

## Is this a breaking change?
- [ ] Yes
- [x] No

## I certify that...
- [x] All relevant unit and integration tests have passed and/or have been updated according to this change.

## Generative AI (GAI) Usage Disclosure
[Reference: Cisco GAI Coding Guidelines](https://cisco.sharepoint.com/sites/AIML/SitePages/Practice-Guide---GAI-Coding.aspx)

- [ ] **Cisco approved GAI tool/IDE was used for coding** - eg: VSCode, Cursor, Windsurf, Codex, Claude code
- [x] **No GAI** - Code was written entirely manually without GAI assistance

**If a GAI tool/IDE was used then select the category that best describes GAI usage in this PR:**
- [ ] **Manual Draft with GAI Refinement** - I/we created a draft manually that was non-substantively modified by GAI (e.g., refactoring was performed by GAI on manually written code) *(Default)*
- [ ] **GAI Draft with Manual Customization** - GAI tool was used to create a draft that I/we subsequently customized or modified.
- [ ] **GAI Generated Code** - Code was generated entirely by GAI

**Additional GAI Usage Details (Optional):**

### DIFF
--- a/src/peer-connection.ts
+++ b/src/peer-connection.ts
@@ -173,6 +173,15 @@ class PeerConnection extends EventEmitter<PeerConnectionEventHandlers> {
   }
 
   /**
+   * Sets the current configuration of the underlying RTCPeerConnection.
+   *
+   * @param configuration - An RTCConfiguration object which provides the options to be set.
+   */
+  setConfiguration(configuration: RTCConfiguration): void {
+    this.pc.setConfiguration(configuration);
+  }
+
+  /**
    * Adds a new media track to the set of tracks which will be transmitted to the other peer.
    *
    * @param track - A MediaStreamTrack object representing the media track to add to the peer connection.
--- a/src/mocks/rtc-peer-connection-stub.ts
+++ b/src/mocks/rtc-peer-connection-stub.ts
@@ -27,6 +27,7 @@ class RTCPeerConnectionStub {
   ): Promise<void> {
     return new Promise(() => {});
   }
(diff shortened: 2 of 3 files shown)


src/peer-connection.ts
```diff
@@ -172,6 +172,15 @@ class PeerConnection extends EventEmitter<PeerConnectionEventHandlers> {
...
   }
 
+  /**
+   * Sets the current configuration of the underlying RTCPeerConnection.
+   *
+   * @param configuration - An RTCConfiguration object which provides the options to be set.
+   */
+  setConfiguration(configuration: RTCConfiguration): void {
+    this.pc.setConfiguration(configuration);
+  }
```
antsukanova, line 182: Could this config be set incorrectly, so we need to catch any issues here with it?

k-wasniowski: Not really, i believe setConfiguration itself won't throw unless it will get appropriate config fields. We are using type RTCConfiguration which is direct type used by `setConfiguration`. But even if, i believe it would be better to delegate an error to the user instead of handling it in some way here silently

antsukanova (approved):

